PR #17, opened 2013-05-13
Reference values in error messages. Fixes #4.

### LINKED ISSUE #4: Reference values in error messages.
It would be good if the user-provided values could be referenced in error messages:

```
schema = Schema( {'<file>': [And(os.path.isfile, error='File "{value}" does not exist')]
                                  ...
```

Rogdham:


skorokithakis:
Could you take a look at #107 as well? I think it relates closely to this PR.

devou87:
> For the removal of `error=e`, there are different reasons for each of them (lines numbers are the one after my commit):
> 
> **line 115**: my bad, I shouldn't have remove that one; I'm changing my commit
> 
> **line 125**: completely useless, since the exception is caught line 132, and is forgotten afterwards; I've added a comment to reflect that
> 
> **line 127**: this one is more difficult to analyse. The exception is caught but re-raised, and the errors get added on line 141.
> If we don't provide an error message, it's ok and normal (behaviour not changed by this PR):
> 
> ```
> >>> Schema({int:str}).validate({4:2})
> Traceback (most recent call last):
> …
> schema.SchemaError: key 4 is required
> 2 should be instance of <type 'str'>
> ```
> 
> If we provide an error message without the `{value}` (e.g. previous behaviour before this PR), everything went fine as well (and I'm explaining why in the next case) (behaviour not changed by this PR):
> 
> ```
> >>> Schema({int:str}, error='Error message').validate({4:2})
> Traceback (most recent call last):
> …
> schema.SchemaError: Error message
> ```
> 
> But if we provide an error message with a `{value}`, it's not as we probably want it to be (behaviour as **before** this PR):
> 
> ```
> >>> Schema({int:str}, error='Error value: {value}').validate({4:2})
> Traceback (most recent call last):
> …
> schema.SchemaError: Error value: {4: 2}
> Error value: 2
> ```
> 
> As you can see, it's getting reported twice. For real errors messages, it's probably not what we want, as it requires modification afterwards. The fact is that in the previous case (without `{value}`), it was reported twice too, but we had a duplicate removal on line 27 witch took care about that.
> 
> Note that it is always possible to have both reports, with `{value}`, if we specify the internal error message (behaviour not changed by this PR):
> 
> ```
> >>> Schema({int:Use(int, error='Internal error: {value}')}, error='External error: {value}').validate({4:'a'})
> Traceback (most recent call last):
> …
> schema.SchemaError: External error: {4: 'a'}
> Internal error: a
> ```
> 
> So what this removal of `error=**

PR #42, opened 2014-04-25
Improve custom error handling.

chadrik:
The `Schema` 'error' kwarg now supports a callable, which returns the error message as a string. Its arguments are:
- `schema`: the underlying schema object
- `data`: the data being validated
- `reason`: a string constant for the type of error encountered
- `details`: further details about the reason for the failure.  may be an exeption, a string, or None

Additionally, error strings passed to the `Schema` 'error' kwarg are now provided more context about the error in the form of four `{fields}` for use with "new-style" string formatting.  These are string representations of the four arguments described above.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -1,5 +1,36 @@
 __version__ = '0.3.0'
 
+error_messages = {
+    'did not validate': '{schema} {reason} {data}',
+    'callable raised exception': '{schema}({data}) raised {details}',
+    'validate raised exception': '{schema}.validate({data}) raised {details}',
+    'invalid value for key': '{reason} {details}',
+    'missed keys': '{reason} {details}',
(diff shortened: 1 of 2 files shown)


schema.py
```diff
@@ -1,5 +1,36 @@
 __version__ = '0.3.0'
 
+error_messages = {
+    'did not validate': '{schema} {reason} {data}',
```
erikrose: I feel like these should be different subclasses of SchemaError. Then we could catch them selectively with try/except in calling code. What do you think?

chadrik: I think that I made this pull request over a year ago and it's obvious that this project is not actively maintained.  I've moved on.  [schemtatics](https://github.com/schematics/schematics) looks interesting....

PR #45, opened 2014-06-19
added nested prioritisation for dict validation

nerdvegas:
Consider the schema:

```
s = Schema({Optional('foo'): int, Optional(basestring): object})
```
Here we would like to have dict validation that ensures 'foo' is integer, but also allows any other extra, unvalidated items. It doesn't work though, because both schema keys have the same priority, and a 'foo' dict entry can be consumed by the Optional(basestring) schema.

This pull request contains a simple change so that prioritisation takes schema contents into account. For example, since a string instance would be used for validation before basestring (non-types before types), it follows that Optional(string instance) now takes precedence over Optional(basestring).

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -80,7 +80,10 @@ def priority(s):
     if type(s) is dict:
         return [5]
     if hasattr(s, 'validate'):
-        return [4] + priority(s._schema)
+        p = [4]
+        if hasattr(s, "_schema"):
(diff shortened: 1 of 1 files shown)


keleshev:
https://github.com/erikrose/schema/commit/e36550d3f027bd5c15af194b397c0a546bd84f7b

PR #46, opened 2014-08-03
I think that‘s more straight forward

beckha:
I think that‘s more straight forward


skorokithakis:
Merged elsewhere, thanks!

Issue #23, opened 2013-09-13
Wrong priority for Optional param


apieceofredcloth:
```
>>> schema = Schema({Optional('id'): Use(int), str: str})
>>> schema.validate({'id': '23', 'other': 'asf'})

>>> {'id': '23', 'other': 'asf'}
```

I think value of id should be converted to integer 23 instead of string '23'("by Use(int)").But "str: str" has a higher priority than "Optional('id')",  so it just ignores the "Use(int)" statement.
I check the code about priority

```
def priority(s):
    if type(s) in (list, tuple, set, frozenset):
        return 6
    if type(s) is dict:
        return 5
    if hasattr(s, 'validate'):
        return 4
    if type(s) is type:
        return 3
    if callable(s):
        return 2
    else:
        return 1
```

Optional has a 'validate' attribute which is inherited from Schema class, so it return 4 here, higher than 3 of the str case.

alefend:
I also have the same issue. I'm trying to validate some query arguments from a web request:
```python
s = Schema({
        Optional('start'): Use(int),
        basestring: object
    })

s.validate({'start': '1'})
{'start': '1'}   # matched basestring: object
```
`'1'` is not converted to `int` because `basestring` matches before `Optional('start')`

As already pointed out by @apieceofredcloth, I would suggest to increase the priority of the `hasattr(s, 'validate')` case.

Another solution may be to have the classes derived from `Schema` (such as `Optional`) expose the same priority of the underlying schema, so `Optional('start')` would have a priority of `1`, while `Optional(str)` would have a priority of `3`.

I can create a pull request if needed.

marc2982:
This isn't fully fixed by #52. Here is a unit test that (sometimes) reproduces the problem i'm seeing on my machine. I'm not sure why it doesn't consistently hit?

```
diff --git a/test_schema.py b/test_schema.py
index 4ec9993..9feebf8 100644
--- a/test_schema.py
+++ b/test_schema.py
@@ -154,6 +154,9 @@ def test_dict_optional_keys():
     # Make sure Optionals are favored over types:
     assert Schema({basestring: 1,
                    Optional('b'): 2}).validate({'a': 1, 'b': 2}) == {'a': 1, 'b': 2}
+    assert Schema({str: {
+        Optional('a', default=False): bool, Optional(str): object}}).\
+            validate({'c': {'a': True, 'b': 2}}) == {'c': {'a': True, 'b': 2}}
```

Test output:
```
[...]

PR #52, opened 2015-02-27
Support defaults via Optional

erikrose:
This lets us insert default values for missing dict keys. I independently arrived at this API before seeing that it's also a leading contender in https://github.com/keleshev/schema/pull/19, so it's at least unsurprising for 2 people. :-)

I had to do a minor twiddle of prioritization so that more generic keys like `str` wouldn't eat the Optionals' lunch, but that shouldn't affect any conceivable non-crashing use case. (See one of the commit messages for details.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -225,6 +225,19 @@ You can mark a key as optional as follows:
     ...         Optional('occupation'): str}).validate({'name': 'Sam'})
     {'name': 'Sam'}
 
+``Optional`` keys can also carry a ``default``, to be used when no key in the
+data matches:
+
+.. code:: python
+
+    >>> from schema import Optional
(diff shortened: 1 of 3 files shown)


erikrose:
The reason the defaults are used verbatim, rather than being passed through any value validators, is that it's more convenient in my use cases to come up with a programmatic value than a raw representation. You can always get the opposite behavior by factoring out any validators in question and manually passing a raw value through them. Inverting the opposite behavior would not always be possible.

erikrose:
I believe this also fixes #23.

keleshev:
Thanks, Erik!

erikrose:
Thanks for the quick merge! :-D

PR #54, opened 2015-02-27
Prioritize Optionals based on one level of depth traversal. Close #44, #45.

### LINKED ISSUE #44: Wrong behaviour with Optional(basestring): object
```python 
> Schema({Optional('attr1'): 'foo', Optional(basestring): object}).validate({'attr1': 'new_foo'})
{'attr1': 'new_foo'}
```

My use case: any keys should be accepted in the dict. Only condition, if I have `attr1` it must be  `foo`.

erikrose:
This enables the common pattern in which there's a 'foo' key with a default and we also take arbitrary other, optional str keys.

We turn priority() into a delegating "generic function", like len().

This lays the groundwork to go deeper--with any "flavor", not just Optionals. First, we'd want to refactor Schema.validate() so it constructs actual Schema subclasses all the time rather than handling many cases inline. For sets, for example, it would construct an Iterable, and Iterable.priority() would know how to tunnel one level inside itself to return a compound priority. Then we could prioritize Optional((1, 0)) over Optional((int,)). The behavior of And() and Or() (and iterables) are up for grabs: perhaps go with the priority of the most specific item inside?

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -77,19 +77,29 @@ COMPARABLE, CALLABLE, VALIDATOR, TYPE, DICT, ITERABLE = range(6)
 
 
 def priority(s):
-    """Return priority for a given object."""
+    """Return the priority with which a schema should be tried against an
+    object.
+
+    A priority is a tuple of at least 1 "flavor". Sorted lexically, they
+    determine the order in which to try schemata against data.
+
+    """
     if type(s) in (list, tuple, set, frozenset):
-        return ITERABLE
-    if type(s) is dict:
-        return DICT
-    if issubclass(type(s), type):
-        return TYPE
-    if hasattr(s, 'validate'):
-        return VALIDATOR
-    if callable(s):
-        return CALLABLE
+        ret = ITERABLE
+    elif type(s) is dict:
+        ret = DICT
+    elif issubclass(type(s), type):
+        ret = TYPE
+    elif hasattr(s, 'validate'):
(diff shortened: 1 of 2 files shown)


erikrose:
Here's one more for you, Vladimir. What do you think? I'm rather enthused to do the mentioned refactoring as well, if you like.

keleshev:
Love the idea of including `priority` into implicit `Validatable` interface. Here's what I was thinking recently: `validate` is a huge type-conditional on many types and priority is a huge type-conditional—that sounds like a great place for dynamic dispatch. How about: `Schema` constructor walks our schema DSL structure and constructs a proper schema with converting:

* `str` -> `Type(str)`
* `[a]` -> `Iterable([a])`
* `a` -> `Literal(a)`

With each of the `Type, Iterable, Literal` having appropriate `validate` method and `priority` property. We'll be able to get rid of a `priority` function, and split the huge `Schema.validate` into many smaller methods like `Type.validate`. And custom user classes will be able to specify their own priority. At that point we will be able to easily compose priorities, we could say that Optional has priority of it's child, which in turn *might* compute or pick priority from it's children. 

What do you think?

keleshev:
Not to forget, but a minor concern: if priority becomes part of public interface, then we should see if there's anything better than `int` for it.

keleshev:
Oh, my! I should have read your #55 first :-)

Issue #64, opened 2015-07-10
Feature Request: Named parts


kosz85:
Hi,
Sometimes I want to only validate some data, and sometimes also process it. It would be nice to have it in one schema. 
Simple Example:
```
>>> schema = Schema({'user': And(int, Use(get_user_from_db, name='import'))})
>>> s.validate({'user': 1})  # current behavior
{'user': <User: username>}
>>> s.validate({'user': 1}, exclude=('import',))
{'user': 1}
```
The second option, validates it, but do not run particular elements, it could be `exclude` and `only` parameters. Not named would be always run. I imagine it could work also like this `Name('import', Use(get_user_from_db))` or with optional multiple names `Name(('import', 'process'), Use(...))`

Issue #63, opened 2015-07-01
Feature Request: Docstrings


elrond79:
I'd love to see support for docstrings/helpstrings/descriptions of some sort.  Not only would this encourage documentation of the schema, but it would allow automated generation of things like argument parsers, and allow us to finally unite all our schema-definition in one place.

keleshev:
Could you elaborate and give an example?

elrond79:
Well, haven't given a lot of thought yet to the exact interface, but maybe something like this:

dino_schema = Schema({
    "Claws": Doc(int, "the number of claws on each hand"),
    "Diet": Doc(Or("meat", "plants", "people"), "what it eats"),
})

The Doc wrappers would be discarded when validating, but would be queryable somehow - given the arbitrary nature schema data, how exactly is a bit of an open question.  Maybe we could do:

>>> dino_schema.doc()
{"Claws": "the number of claws on each hand", "Diet": "what it eats"}
>>> dino_schema.doc()["Claws"]
"the number of claws on each hand"
>>> dino_schema._schema["Diet"].doc()
"what it eats"

kosz85:
You could do that with #64. 
If you assume that we will define restricted name `__doc__` then:
`dino_schema.validate(only='__doc__')`
 could behave like you described, and for other set of namespaces do nothing.

PR #65, opened 2015-07-10
Made schema more extendable with simple trick. Issue #63 #64 

kosz85:
I made feature request #64 then read issue #63, and after some thought and seeing code found out that it's ultra easy to achieve. I don't think #64 and #63 need to be implemented, but it's nice to have that possibility without modifying base schema code.
Doc class from #63 is implemented as a test `test_validate_kwargs_doc_example`, to show how easy it is.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -35,19 +35,19 @@ class And(object):
         return '%s(%s)' % (self.__class__.__name__,
                            ', '.join(repr(a) for a in self._args))
 
-    def validate(self, data):
+    def validate(self, data, **kwargs):
         for s in [Schema(s, error=self._error) for s in self._args]:
-            data = s.validate(data)
+            data = s.validate(data, **kwargs)
         return data
 
 
 class Or(And):
 
-    def validate(self, data):
+    def validate(self, data, **kwargs):
         x = SchemaError([], [])
         for s in [Schema(s, error=self._error) for s in self._args]:
             try:
-                return s.validate(data)
+                return s.validate(data, **kwargs)
             except SchemaError as _x:
                 x = _x
         raise SchemaError(['%r did not validate %r' % (self, data)] + x.autos,
@@ -64,7 +64,7 @@ class Use(object):
     def __repr__(self):
         return '%s(%r)' % (self.__class__.__name__, self._callable)
 
-    def validate(self, data):
+    def validate(self, data, **kwargs):
         try:
             return self._callable(data)
         except SchemaError as x:
@@ -102,16 +102,16 @@ class Schema(object):
     def __repr__(self):
         return '%s(%r)' % (self.__class__.__name__, self._schema)
 
-    def validate(self, data):
+    def validate(self, data, **kwargs):
         s = self._schema
         e = self._error
         flavor = priority(s)
         if flavor == ITERABLE:
-            data = Schema(type(s), error=e).validate(data)
+            data = Schema(type(s), error=e).validate(data, **kwargs)
             o = Or(*s, error=e)
-            return type(data)(o.validate(d) for d in data)
+            return type(data)(o.validate(d, **kwargs) for d in data)
         if flavor == DICT:
-            data = Schema(dict, error=e).validate(data)
+            data = Schema(dict, error=e).validate(data, **kwargs)
             new = type(data)()  # new - is a dict of the validated values
             x = None
             coverage = set()  # matched schema keys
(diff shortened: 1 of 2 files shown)


skorokithakis:
This looks good to me, it's backwards compatible and doesn't hurt anything. What do you think, @keleshev ?

keleshev:
When would you use something like this? I could see how that works, but I don't understand the use-case.

skorokithakis:
It's for passing data into your subclasses. It's generally good design for `validate()` to pass down extra data, as validation on subclasses may want to use it.

kosz85:
You can easily extend schema, and made it even more powerful, import example from my project:
```
class Import(object):
    """
    >>> Import(Use(lambda x: 'V'), Use(lambda x: 'I')).validate(1)
    'V'
    >>> Import(Use(lambda x: 'V'), Use(lambda x: 'I')).validate(1, importing=True)
    'I'
    """
    def __init__(self, validation, processing):
        self.validation = validation
        self.processing = processing

    def validate(self, data, **kwargs):
        if kwargs.get('importing', False):
            return self.processing.validate(data, **kwargs)
        return self.validation.validate(data, **kwargs)
```
then you define:
```
schema = {
"some_field": And(
        Or(int, LazyType(lazy_model)),
        Import(
            Use(get_object_pk),
            Use(get_object_instance(lazy_model))
        ))
}
```
And use:
- to validate: `Schema(data).validate()`
- to import data `Schema(data).validate(importing=True)`

I have in the same style auto documentation. I use it along with Django, so validate do also exporting for me, and `importing` is changing all ids into objects. So kwargs do the magic for me with very nice interfejs :)

kosz85:
I merged it due to conflicts again.

kosz85:
Please review and merge, I resolved conflicts again

skorokithakis:
This seems pretty straightforward and backwards-compatible, any objections from anyone? @keleshev?

keleshev:
I carefully read code and example and still don't understand—what is an end goal? Could you make a user story for this? Maybe write some docs? 

I suggest to consider this feature after releasing the current version.

kosz85:
@keleshev I give you example few posts ago. You can do things like class `Import`, so you can create your own set of extensions to Schema on top of Schema without editing any part of it. That's powerful, you can  control your `validate` methods using `**kwargs` parameters.

There is nothing really hard, check my example (especially doc tests) and see that output changes when you put different kwarg. I'm using exactly smth like this to serialize and unserialize django models for my planned in time  notifications arguments (to not store models) and in the same time having type validations for them! That's just great!
In my case I push real objects and they all are converted to id's when using `schema.validate(data)` and convert them back to objects few months later when using `validate(data, import=True)`.

keleshev:
@kosz85 I have seen bot the `Import` and `Docs` examples, I just couldn't understand their use cases. I now vaguely understand `Import`'s. How about this instead:

```python
# with #65
schema = Import(Use(check_pk), Use(get_instance))
json = schema.validate(json)  # validate only
mole = schema.validate(json, importing=True)  # validate and load

# without #65
schema = lambda importing=False: Use(get_instance if importing else check_pk)
json = schema().validate(json)  # validate only
mole = schema(importing=True).validate(json)  # validate and load
```

kosz85:
Your example is more like using two different schemas, and is not passing kwargs down the objects. You can't do smth like this without declaring two schemas:
```
schema({
    'smth': {
          'other': Import(Use(check_pk), Use(get_instance))
    }
})
```
I build one schema with functions like `schema_for_model('app.Model')`, and it's only one clean definition without lambda and function hacks. I also don't need to know all supported kwargs when defining schemas. So in future extending it with another kwarg is easier.
Check this:
```
def schema_for_model(model):
    if isinstance(model, basestring):
        lazy_model = lambda: get_model(*model.rsplit('.', 1))
    else:
        lazy_model = lambda: model
    return And(
        Or(int, LazyType(lazy_model)),
        Import(
            Use(get_object_pk),
            Use(get_object_instance(lazy_model))
        ))
```
I don't care what kwargs it support, I just need to use it like this:
```
schema = Schema({
     'user': schema_for_model('auth_app.User'),
     'example': schema_for_model('examples.ExampleModel'),
})
```
If I need i can later add smth new in my schema utils, and use it like this:
```
schema.validate(import=True)
schema.validate(docs=True)
schema.validate(json_schema=True)
```
And I really don't need to change any place where I already use it, just few classes that are defining it.

I hope you will merge it, if not then it's ok, I already use at production my fork anyway.

keleshev:
Sorry, not convinced, I think this is a scope creep for **schema**. **Schema** is fundamentally based on the idea of functions and objects composition, unlike for example JSON schema spec. See how many `lambda`s are there in README. Also, personally dislike `**kwargs` unless all invalid keys are reported.

kosz85:
Your project, your choice, but personally I think you are doing a big mistake. As a last resort to convince you try to implement solution for my problem. Add smth that will load complicated schemas (not smth flat that you mimic with simple lambda), with importing objects  in lazy way resolving models, and  then add  some new key argument, for example `json_export`, which will convert all not base types like deatetitme to json acceptable format. In my  solution it takes writing one  class and use it in 3 more places, in your it would be the number of schemas you have.
All bigger projects use  `**kwargs` there is nothing to be afraid.

PR #67, opened 2015-09-11
Unique exceptions

werehuman:
Includes #66 

Some error messages aren't looks user friendly. For example, when some keys are missing you'll see `missing keys set(['foo', 'bar'])`. Discussions about which message is the most beautiful may be infinite, so I propose another solution.

Each error message can be easily overrided. Accepted overriding by subclassing and by passing new handlers into constructor.

Upd: Now error messages in master are good enough, but this PR still useful for l10n and other customizations.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -111,13 +111,20 @@ def _priority(s):
 
 class Schema(object):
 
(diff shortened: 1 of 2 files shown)


werehuman:
Rebased into master.

PR #74, opened 2015-09-22
Simplify uniq.

skorokithakis:


schema.py
```diff
@@ -13,9 +13,7 @@ def __init__(self, autos, errors):
     @property
     def code(self):
         def uniq(seq):
-            seen = set()
-            seen_add = seen.add
-            return [x for x in seq if x not in seen and not seen_add(x)]
+            return list(set(seq))
```
guilload: I don't know if it matters but the ordering might be lost by doing so.

skorokithakis: You're right, I'm getting errors going this so I'll hold off. I'll add a comment saying order will be lost instead, thank you.

keleshev: A better name for this would be `stable_unique`, then the confusion would not arouse.

PR #89, opened 2015-09-27
Travis/Tox: Add testing with Python3.5

sjakobi:


sjakobi:
I've added a change that should keep tox from recreating the various venvs every time.

PR #90, opened 2015-09-29
bump version to 0.3.2.dev0 as 0.3.1 was released long ago

jab:
Currently when you install the latest development version via
e.g. pip install 'git+https://github.com/keleshev/schema.git'
you end up with a schema package that still advertises itself
as version 0.3.1, even though it's a later version.

Bumping the version to 0.3.2.dev0 makes it easier for users
to confirm whether a bug released in 0.3.1 has since been
fixed in master, as we can now inspect `schema.__version__`
and get a different value than the version with the bug.

I learned it's good practice to bump the version like this as the
next commit to master immediately after doing a release.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -1,4 +1,4 @@
-__version__ = '0.3.1'
+__version__ = '0.3.2.dev0'
 
 
 class SchemaError(Exception):


skorokithakis:
This is a good idea, but I'm not crazy about the choice of version, as it breaks [semver](http://semver.org/). Can you change it to `0.3.2-dev0` instead?

jab:
Sadly the choice of version is not mine, it's straight out of [PEP 0440](https://www.python.org/dev/peps/pep-0440/#public-version-identifiers), which trumps semver in Pythonland. Setuptools will either complain or convert to what I used.

jab:
(Whoops, linked to the wrong section of the PEP; edited my previous comment to fix)

skorokithakis:
Aw, OK.

PR #101, opened 2016-02-12
add strict flag in Schema class to skip wrong key validation without wildcard

drgarcia1986:
To avoid `SchemaError` for validations with more keys of the schema:

```python
>>> Schema({'key', 'value'}, strict=False).validate({'key', 'value', 'foo': 'bar})
>>> {'key', 'value'}
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -257,6 +257,19 @@ for the same data:
     >>> Schema(And(Or(int, float), lambda x: x > 0)).validate(3.1415)
     3.1415
 
+Extra Keys
+~~~~~~~~~~
+
+The ``Schema(...)`` parameter ``ignore_extra_keys`` do not validate extra keys and also do not return them after validation.
+
+.. code:: python
+
+    >>> schema = Schema({'name': str}, ignore_extra_keys=True)
+    >>> schema.validate({'name': 'Sam', 'age': '42'})
+    {'name': 'Sam'}
+
+Otherwise, any extra key raise a ``SchemaError``.
+
 User-friendly error reporting
 -------------------------------------------------------------------------------
 
--- a/README.rst
+++ b/README.rst
@@ -260,7 +260,7 @@ for the same data:
 Extra Keys
 ~~~~~~~~~~
 
-The ``Schema(...)`` parameter ``ignore_extra_keys`` do not validate extra keys and also do not return them after validation.
+The ``Schema(...)`` parameter ``ignore_extra_keys`` causes validation to ignore extra keys in a dictionary, and also to not return them after validating.
 
 .. code:: python
 
@@ -268,7 +268,8 @@ The ``Schema(...)`` parameter ``ignore_extra_keys`` do not validate extra keys a
     >>> schema.validate({'name': 'Sam', 'age': '42'})
     {'name': 'Sam'}
 
-Otherwise, any extra key raise a ``SchemaError``.
+If you would like any extra keys returned, use ``object: object`` as one of the key/value pairs, which will match any key and any value.
+Otherwise, extra keys will raise a ``SchemaError``.
 
 User-friendly error reporting
 -------------------------------------------------------------------------------
--- a/schema.py
+++ b/schema.py
@@ -96,10 +96,10 @@ def _priority(s):
 
 class Schema(object):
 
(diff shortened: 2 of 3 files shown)


gabrielhpugliese:
I talked with @drgarcia1986 about this issue and it's useful when you don't have a consistent data input and don't want to explode things in production.

drgarcia1986:
@skorokithakis you can review ?

skorokithakis:
I have a small objection about the API. I currently do:

~~~
Schema({'key', 'value', object: object}, strict=False).validate({'key', 'value', 'foo': 'bar})
~~~

as a catch-all, which is explicit about the fact that the resulting structure will include those keys. What worries me about `strict=False` is that the keys are omitted, although it may be a good addition to the above alternative, as you may not want the keys back but not want validation to fail either.

skorokithakis:
Also, in the "wrong keys" case, does the code actually mean "superfluous keys"? Or can execution reach that point for a case other than extra keys in the dict?

drgarcia1986:
I don't see any other case to trigger the "wrong keys" case :neutral_face:

mianos:
I like the strict=False, but I'd want it to return all the unparsed keys unchanged along with the checked keys converted, as it does now.
In quite a lot of my code I do a check in the json receiver and dispatch to a specialised handler that processed the json with another, more specific schema. 

If it returned the unchecked keys unchanged you can just run check another schema later, with the keys they were ignored earlier ignored, passing the already converted data types.

skorokithakis:
Returning the unchecked keys unchanged is already possible with {object: object}, though.

mianos:
Thanks for that. I didn't realise it was so obvious, using an 'object' on the left to catch all possible keys. (Embarassingly didn't read your comment above properly either). It works perfectly but it is a little scary.

drgarcia1986:
Catch all possible keys is another case, I think, maybe we add another parameter for this situation.

gabrielhpugliese:
Hey guys, it would be nice if we decide what we should do. @skorokithakis will it be merged? :)

skorokithakis:
Sure, if we decide on the API. I'm liking this, since we can always have Schema return other keys with the `object: object` trick. I'm not a fan of the name `strict`, I'd prefer if it was called something like `ignore_extra_keys`, or something like that.

By the way, @drgarcia1986, can you update the docs with the parameter so the PR is more complete?

drgarcia1986:
Update parameter name.
c/c @skorokithakis

drgarcia1986:
now add documentation :smile:

skorokithakis:
Looks good to me, I think I can merge after a very minor correction to the docs.

README.rst
```diff
@@ -257,6 +257,19 @@ for the same data:
     >>> Schema(And(Or(int, float), lambda x: x > 0)).validate(3.1415)
     3.1415
 
+Extra Keys
+~~~~~~~~~~
+
+The ``Schema(...)`` parameter ``ignore_extra_keys`` do not validate extra keys and also do not return them after validation.
```
skorokithakis: Better wording:

> The ``Schema(...)`` parameter ``ignore_extra_keys`` causes validation to ignore extra keys in a dictionary, and also to not return them after validating. If you would like any extra keys returned, use ``object: object`` as one of the key/value pairs, which will match any key and any value.

diegogarcia-luizalabs: Great

drgarcia1986: Great

drgarcia1986:
```
The ``Schema(...)`` parameter ``ignore_extra_keys`` causes validation to ignore extra keys in a dictionary, and also to not return them after validating. 

.. code:: python

    >>> schema = Schema({'name': str}, ignore_extra_keys=True)
    >>> schema.validate({'name': 'Sam', 'age': '42'})
    {'name': 'Sam'}

If you would like any extra keys returned, use ``object: object`` as one of the key/value pairs, which will match any key and any value.
Otherwise, any extra key raise a ``SchemaError``.
```
?

skorokithakis: "Otherwise, extra keys will raise a ``SchemaError``."

Other than that, perfect!

drgarcia1986: Done :+1:

skorokithakis:
Alright, looks good, thanks!

Issue #98, opened 2015-11-06
Allow formatting for custom error messages


MaxNoe:
I would be hugely helpfull, if one could format the information
which is available for the stanard error messages also into the custom errors.

Idea:

```
from schema import Schema
from os import path

test_existance = Schema(path.exists, error='File "%(data)r" not found!')
```

skorokithakis:
Hmm, what would `data` be in this case? Are you proposing that Schema runs `error.format(data)` before display?

MaxNoe:
Yes. Currently the code for the build in message is:
For example for functions:
```
'%s(%r) should evaluate to True' % (f, data)
```
Or for Use:
```
'%s(%r) raised %r' % (f, data, x)
```
It would be nice if the custom error messages could also use `f`,`data` or `x`.

bernalrs:
I guess we can now close this issue ?

PR #107, opened 2016-06-22
Allow the error parameter string to contain string formatting parameters

skorokithakis:
This PR adds a call to format on the `error` parameter before displaying it, passing the data into the formatter. This allows custom error messages to change from something like `"Invalid user id."` to `"Invalid user id: {}"`, which would display as `"Invalid user id: 9243"`.

Please let me know if I missed something somewhere.


skorokithakis:
This PR relates to issue #98.

PR #108, opened 2016-07-12
Adding regular expression (regex) support

gusmonod:
Hi,

Here is a proposition to add regex support to the library, with the use of `re.pattern.search` method from the Python core regex library (support was tested on python 2.6.9, 2.7.12, 3.3.0, 3.4.3, 3.5.2, pypy-5.3 and pypy3-2.4.0).

Four simple test cases were added as well.

Note that support for python 3.2 was not tested because of an issue with tox and virtualenv: https://github.com/travis-ci/travis-ci/issues/5517

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,8 +117,25 @@ this method on corresponding data as ``data = obj.validate(data)``. This method
 may raise ``SchemaError`` exception, which will tell ``Schema`` that that piece
 of data is invalid, otherwise—it will continue validating.
 
-As example, you can use ``Use`` for creating such objects. ``Use`` helps to use
-a function or type to convert a value while validating it:
+An example of "validatable" is ``Regex``, that tries to match a string or a
+buffer with the given regular expression (itself as a string, buffer or
+compiled regex ``SRE_Pattern``).
+
+.. code:: python
+
+    >>> from schema import Regex
+    >>> import re
+
+    >>> Regex(r'^foo').validate('foobar')
+    'foobar'
+
+    >>> Regex(r'^[A-Z]+$', flags=re.I).validate('those-dashes-dont-match')
+    Traceback (most recent call last):
+    ...
+    SchemaError: Regex('^[A-Z]+$', flags=re.IGNORECASE) does not match 'those-dashes-dont-match'
+
+For a more general case, you can use ``Use`` for creating such objects.
+``Use`` helps to use a function or type to convert a value while validating it:
 
 .. code:: python
 
@@ -267,8 +284,8 @@ User-friendly error reporting
 -------------------------------------------------------------------------------
 
 You can pass a keyword argument ``error`` to any of validatable classes
-(such as ``Schema``, ``And``, ``Or``, ``Use``) to report this error instead of
-a built-in one.
+(such as ``Schema``, ``And``, ``Or``, ``Regex``, ``Use``) to report this error
+instead of a built-in one.
 
 .. code:: python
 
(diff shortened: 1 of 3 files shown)


skorokithakis:
This looks great, thank you. Could you add a few more, varied test cases, like `{"myregex": Regex("foo")}` and things like that? Just to make sure that it doesn't break with some other obvious uses.

skorokithakis:
Thank you! Is this good to merge?

gusmonod:
I found a few other bugs when adding new tests, please wait a second ;-)

skorokithakis:
No problem, I'll merge when you're ready (that's why I asked for more tests :smile:).

gusmonod:
Ok, I have now added a few more tests, and made a minor design choice that we probably want to discuss:

Instead of asserting the pattern is string-like (buffer interface) and raising a `SchemaError` when the data to test isn't string-like, I decided to let the standard `TypeError` be raised so that the error message is a bit more explicit.

What do you think of this approach?

Note that I also wanted to check for this kind of errors in the tests, but wasn't able to find an obvious type that is string-like yet isn't a string. Any ideas on that?

skorokithakis:
I may be misunderstanding, in which case correct me if I'm wrong, but from what I understood:

In general, I am against letting exceptions be raised in the validation. If the object to be matched (not the pattern, the supplied object) is not string-like, validation should fail with a `SchemaError`. That's because, if I expect a pattern that matches `"hello"` and the user supplies `None`, I want validation to fail as normal. In general, while validating, I don't care about specifically what the user supplied, I only care if it validates or not. Therefore, I would prefer that you not let the `TypeError` be raised, but at least re-raise the error message with a `SchemaError`.

As for the tests, I don't think that searching for a string-like type that's not a string is necessary, unless you have a specific case you want to test. If it's obscure enough that you're having trouble finding a type to test it, you should probably let it fail :smile:. In practice, it will either work or it will raise a `SchemaError`, which is the right behaviour anyway.

gusmonod:
Right, so from what you say, I think I did the right thing 😄.

I am raising a `TypeError` when the **pattern** is incorrect, and catching the `TypeError` to replace it with a `SchemaError` if the **data** to match against the pattern is incorrect.

gusmonod:
OK, I have now finally added the last bits of the `Regex` tests and documentation.

Note that I have also added a few lines of code to the implementation so that the error output is slightly nicer is Python versions < 3.4 (<3).

skorokithakis:
Yep, that's exactly correct, thank you! Let me know when you're ready to review.

gusmonod:
Ready right now!

skorokithakis:
Ah, I appear to have introduced a small conflict when changing the version...

gusmonod:
I just rebased my pull request in order to fix the conflict introduced by 6780c25ed04ed6eae74571d91a761153272e2da4 and b976cb7044c5c78dbe26d6217f6de654f03c7916.

As for your nitpick, I am completely fine to change that, but then I think it would be weird to keep [this error message](https://github.com/keleshev/schema/blob/b976cb7044c5c78dbe26d6217f6de654f03c7916/schema.py#L187) written the other way.

So in my opinion, we should either have **both** say `"<schema> does not match <data>"` or "<data> does not match <schema>`, but not have two different ways.

skorokithakis:
Hmm, you're right. Let's leave that error message as it is now, I will review all the messages later on and possibly change them. Let me review one last time!

skorokithakis:
Merged, thank you for your help!

gusmonod:
Nice!

gusmonod:
Hi, can you please push this new version to PyPI?

skorokithakis:
Oops, sorry, I didn't see the last comment. Published 0.6.0!

gusmonod:
Thanks!

PR #110, opened 2016-07-27
Fixing failing tests for error formatting feature. (#98)

bernalrs:
Notice that the tests were failing would be good if we could include this feature in the next release.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -69,10 +69,13 @@ class Use(object):
         try:
             return self._callable(data)
         except SchemaError as x:
-            raise SchemaError([None] + x.autos, [self._error.format(data) if self._error else None] + x.errors)
+            raise SchemaError([None] + x.autos,
+                              [self._error.format(data)
+                               if self._error else None] + x.errors)
         except BaseException as x:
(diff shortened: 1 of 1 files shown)


skorokithakis:
This is for when the user hasn't specified an error string, correct?

bernalrs:
Yes that right in some scenarios we are expecting None to be included in the error list it appears this was the intended design however I am not sure about the rationale behind the decision.

skorokithakis:
Hmm, yes, I'm not aware either, @keleshev would know. I'm good to merge this, can you split the long lines a bit so all tests pass?

bernalrs:
All good sorry about the lines length my devenv is set to 120 characters.

skorokithakis:
No problem, thank you!

bernalrs:
Could you please shed light into the process getting this into master and schema's next release ?

skorokithakis:
Sure, it's basically "if you care enough to submit a PR, you should get a quick release". Therefore, [your fix is now in](https://pypi.python.org/pypi/schema/0.6.1). Thanks again!

PR #117, opened 2016-08-04
Sort missing keys

TBoshoven:
Wrong keys are sorted, but for some reason, missing keys are not.
Sorting them in the error message looks better and is useful for testability.


skorokithakis:
Thank you!

Issue #121, opened 2016-09-19
Error when error argument missing


DaMaillard:
Error when serialazing a SchemaError object : 

```python
>>> import copy
>>> from schema import SchemaError
>>> s = SchemaError('a', None)
>>> copy.deepcopy(s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/damienmaillard/.virtualenvs/bricklane/lib/python3.4/copy.py", line 182, in deepcopy
    y = _reconstruct(x, rv, 1, memo)
  File "/Users/damienmaillard/.virtualenvs/bricklane/lib/python3.4/copy.py", line 295, in _reconstruct
    y = callable(*args)
TypeError: __init__() missing 1 required positional argument: 'errors'
```

DaMaillard:
@skorokithakis Can you please update PyPI ?

skorokithakis:
Sure!

skorokithakis:
https://pypi.python.org/pypi/schema/0.6.3

PR #123, opened 2016-09-19
Adding serialization test

DaMaillard:
issue #121


skorokithakis:
Damn, this breaks the tests. Can you fix it?

skorokithakis:
I'm going to have to revert the previous release if this doesn't get fixed within the next half hour, I'm afraid :/

gusmonod:
Here is a simple way to fix it: instead of using an empty list, just use `None` and then let a list be built with `[None]`

gusmonod:
Oh, and @skorokithakis, I think you should also add this to test that there is no regression.

skorokithakis:
I don't understand how this worked (although it may not have). It's missing an import of `copy` and typos `s1` to `s`. I guess it never worked.

PR #125, opened 2016-10-28
Implement Forbidden.

MathieuDuponchelle:
See the documentation and inline comments for more details.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -249,6 +249,37 @@ data matches:
 Defaults are used verbatim, not passed through any validators specified in the
 value.
 
+You can mark a key as forbidden as follows:
+
+.. code:: python
+
+    >>> from schema import Forbidden
+    >>> Schema({Forbidden('age'): object}).validate({'age': 50})
+    Traceback (most recent call last):
+    ...
+    SchemaForbiddenKeyError: Forbidden key encountered: 'age' in {'age': 50}
+
+A few things are worth noting. First, the value paired with the forbidden
+key determines whether it will be rejected:
+
+.. code:: python
+
+    >>> Schema({Forbidden('age'): str, 'age': int}).validate({'age': 50})
+    {'age': 50}
+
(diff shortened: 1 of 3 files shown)


MathieuDuponchelle:
I'm completely open to renaming the class, but unlike some other additions I've made on top of schema, this one seemed generic enough that it could be upstreamed, good name or not :)

skorokithakis:
Hmm, this is an interesting addition, I think I like it. @keleshev, what do you think?

@MathieuDuponchelle, can you fix the failing tests in the mean time?

MathieuDuponchelle:
Done too :)

MathieuDuponchelle:
@skorokithakis , should I rebase this ?

skorokithakis:
Yes please. Looking at it again, I would like to merge it when tests pass, it seems like a very useful feature (especially the fact that you can forbid based on matching values).

I'm wondering, however, whether another way might be more practical, when forbidding values. The semantics seem a bit wrong, as it seems to me that forbidding a value should be closer to `{"key": Forbidden("value")}`. Unfortunately, that would be problematic in allowing other values and with multiple keys, whereas your current proposal's usage is clear enough. Obviously we can't have something like `{"key": int, "key": Forbidden(string)}`. I'm probably nitpicking, and what you've done is fine.

@keleshev, do you want to chime in?

MathieuDuponchelle:
Rebased then :)

I see your point with "value-based forbidding", but I'm not sure there's another more satisfying solution, this one has the advantage of "combinability" with Optional, I think as long as it is explicitly documented it should be fine.

afaict the failing test on travis isn't due to this PR, but some sort of setup issue ?

skorokithakis:
Yeah, I thought I fixed the tests issue, but clearly not. Good enough for me, although I wanted @keleshev to opine on this. I guess we can merge now, thank you for all your help!

PR #126, opened 2016-10-28
Add 'Use' to schema.__all__

MathieuDuponchelle:
I tend to do `from schema import *`, didn't initially
understand why Use didn't want to come :)

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -6,7 +6,7 @@ import re
 
 __version__ = '0.6.5'
 __all__ = ['Schema',
-           'And', 'Or', 'Regex', 'Optional',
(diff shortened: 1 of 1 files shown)


skorokithakis:
Oops, thanks! Although, you're missing a comma, so you broke the package :)

MathieuDuponchelle:
Oops, thanks! :P 

I modified the import in the test to use that form, that way what's in __all__ will also be tested in the future :)

MathieuDuponchelle:
Fixed :)

skorokithakis:
Thanks!

PR #127, opened 2016-12-05
Schema can be inherited

her0e1c1:
Hi nice to meet you!
I love your python library, which is small but powerful 😄 

But I have a case to convert data before validating.
When I tried override `validate` method of `Schema`, I found the `Schema` in `And` and `Or` class need to be changed.

I am happy if you merge this commit or give me some advice 😄 

Thanks!

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -66,8 +66,9 @@ class And(object):
     """
     def __init__(self, *args, **kw):
         self._args = args
-        assert list(kw) in (['error'], [])
+        assert sorted(list(kw)) in (['error', 'schema'], ['error'], [])
         self._error = kw.get('error')
+        self._schema = kw.get('schema', Schema)
 
     def __repr__(self):
         return '%s(%s)' % (self.__class__.__name__,
@@ -80,7 +81,7 @@ class And(object):
         :param data: to be validated with sub defined schemas.
         :return: returns validated data
         """
-        for s in [Schema(s, error=self._error) for s in self._args]:
+        for s in [self._schema(s, error=self._error) for s in self._args]:
             data = s.validate(data)
         return data
 
@@ -96,7 +97,7 @@ class Or(And):
         :return: return validated data if not validation
         """
         x = SchemaError([], [])
-        for s in [Schema(s, error=self._error) for s in self._args]:
+        for s in [self._schema(s, error=self._error) for s in self._args]:
             try:
                 return s.validate(data)
             except SchemaError as _x:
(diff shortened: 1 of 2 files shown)


schema.py
```diff
@@ -66,8 +66,9 @@ class And(object):
     """
     def __init__(self, *args, **kw):
         self._args = args
-        assert list(kw) in (['error'], [])
+        assert sorted(list(kw)) in (['error', 'schema'], ['error'], [])
         self._error = kw.get('error')
+        self._schema = kw.get('schema', Schema)
```
skorokithakis: Hmm, what is the rationale for this? Could you add a short comment explaining why someone can pass the class here? Why shouldn't every subclass do `self.__class__` instead?

her0e1c1: Thanks for your feedback! I think there is a bit problem if you use Schema directly because it's public class but it can not be inherited.

About this code `self.__class__`, I think it's good enough to use it in only `Schema`. I mean about other code, it's good to pass inherited Schema class instead

skorokithakis:
Hmm, I'd like to merge this, but checks are failing. I guess it's just Pypy's problem, and that the PR is fine, is that correct?

her0e1c1:
@skorokithakis I'm glad to here that! So please check it by myself later :D

her0e1c1:
I don't know the reason and it could help you but I ran this command on this branch
```
docker run --rm -it -v `pwd`:/t -w /t pypy:3 bash -c 'pip install tox && tox -e pypy3'
```
I got test success

skorokithakis:
Looks like it's because pip doesn't work with pypy3 (it reports itself as Python 3.2, which pip doesn't work with). Merging, thank you!

dAnjou:
@her0e1c1 When I look at your test I'm wondering whether it would have been easier to simply do this:
```python
from schema import Schema, Use


def convert(data):
    if isinstance(data, int):
        return data + 1
    return data

s = {
    'k': Use(convert),
    'd': {
        'k': Use(convert),
        'l': [{'l': [Use(convert)]}]
    }
}
v = {'k': 1, 'd': {'k': 2, 'l': [{'l': [3, 4, 5]}]}}
d = Schema(s).validate(v)

assert d['k'] == 2
assert d['d']['k'] == 3
assert d['d']['l'][0]['l'] == [4, 5, 6]

```

her0e1c1:
Hi @dAnjou :) Yeah the test case can be more simple.
But my test purpose is to inherit. I mean, if there is not my commits, the test would fail.

And in my case (sorry suddenly), I want to convert sqlalchemy model class to dict before validating.
I think In the case, you can not use `Use` only

PR #131, opened 2017-03-07
Apply ignore flag recursively for dict

hangyan:
#131

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -66,8 +66,9 @@ class And(object):
     """
     def __init__(self, *args, **kw):
         self._args = args
-        assert sorted(list(kw)) in (['error', 'schema'], ['error'], [])
+        assert set(kw).issubset(['error', 'schema', 'ignore_extra_keys'])
         self._error = kw.get('error')
+        self._ignore_extra_keys = kw.get('ignore_extra_keys', False)
         # You can pass your inherited Schema class.
         self._schema = kw.get('schema', Schema)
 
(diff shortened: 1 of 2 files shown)


skorokithakis:
This looks good to me, are there any downsides? Does it change any semantics in undesirable ways?

hangyan:
@skorokithakis I think it's natural for people to think this flag is recursively. I have add the support for `list(dict)` too. There may be one downside, for example: `Schema([1, 0]).validate([1, 1, 0, 1])`, it we add the ignore flag, `Schema([1, 0], ignore_extra_keys=True).validate([1, 1, 0, 1, 2])`, it has no effect for simple types. But that's also make sense, because only `dict` have keys. By the we, the CI test always failed at env `pypy3`, is there something wrong on the env? Can anyone fix it? Thanks!

skorokithakis:
That makes sense, thanks. The `pypy3` test is problematic, yes. The example, `Schema([1, 0], ignore_extra_keys=True).validate([1, 1, 0, 1, 2])`, will fail to validate, correct? It *should* fail to validate, as, as you say, only dicts have keys.

hangyan:
Yes, it will fail, it seems no need to add this check to unittest, so i didn't. I can give a screenshot:

image

skorokithakis:
Great, thank you!

hangyan:
Thanks. (I forgot to update the docs.. will do it as soon as possible!

PR #140, opened 2017-05-04
added __hash__ and __eq__ based on sub-schema to Optional

kurtbrose:


test_schema.py
```diff
@@ -226,6 +226,8 @@ def test_dict_optional_keys():
     # Make sure Optionals are favored over types:
     assert Schema({basestring: 1,
                    Optional('b'): 2}).validate({'a': 1, 'b': 2}) == {'a': 1, 'b': 2}
+    # Make sure Optionals hash based on their key:
+    assert len({Optional('a'): 1, Optional('a'): 1}) == 1
```
skorokithakis: You also need this:

~~~
assert len({Optional('a'): 1, Optional('a'): 1, Optional('b'): 2}) == 2
~~~

PR #141, opened 2017-05-04
added __hash__ and __eq__ based on sub-schema to Optional

kurtbrose:
change to unit test as requested (new PR to keep history clean)


skorokithakis:
Thank you! You don't need a new PR, though, we can just squash on merging.

skorokithakis:
Thanks!

PR #145, opened 2017-09-27
Implement Const validator to keep data unchanged

insideable:
please take a look in `test_schema.py` to see use case example for `Const`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,9 @@ cache:
 install: pip install codecov tox
 
 env:
-  - TOX_ENV=py26
   - TOX_ENV=py27
-  - TOX_ENV=py33
   - TOX_ENV=py34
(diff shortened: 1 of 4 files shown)


skorokithakis:
Thank you, I like this and it seems simple enough. Could you get the tests passing? You can remove the Python3.3 interpreter one, as that's not your fault, but I would like the PEP8 tests to be passing.

insideable:
@skorokithakis I have fixed pep8, but py26, py33 and pypy still failed due these versions of python interpreters are missed. Should I change something in `.travis.yml` ?

skorokithakis:
Yes please, they need to be removed from there.

insideable:
@skorokithakis here we go!

skorokithakis:
Thank you!

PR #146, opened 2018-02-16
Implement Predicate decorator

doloopwhile:
Hello. I added `Predicate` decorator.

Currently, To define a functional schema, We have to write a `def` statement and an assignment statement.

```
def some_condition(x):
   ...
   ...
   (Some validation logic longer than 10 lines)
   ...
   ...

some_condition.__name__ = "Error message" # When define it as naked function
some_condition = Schema(some_condition, error="Error message") # When define it as object
```

`Predicate` make it possible to define a functional schema as one `def` statement.

```
@Predicate(error="Error message")
def some_condition(x):
   ...
   ...
```

I feel later style is more readable (in the same reason of `@getter` defining properties).


skorokithakis:
Thank you for your contribution! Hmm, this looks useful, but I don't know if the verbose way is currently so big a pain to justify including this predicate in the library. On the other hand, it can't hurt, but it does make the library a bit more complicated. You're literally replacing this:

```python
some_condition = Schema(some_condition, error="Error message")
```

with this:

```python
@Predicate(error="Error message")
```

correct? The latter *is* more readable, true, but I'm not sure that's not enough justification to include this. Does anyone else want to chime in?

test_schema.py
```diff
@@ -592,3 +592,25 @@ def validate(self, data):
...
+def test_predicate():
+    @Predicate()
+    def p1(x):
+        return x == 1
+    assert p1._error is None and not p1._ignore_extra_keys
+
+    @Predicate("ERROR2", ignore_extra_keys=True)
+    def p2(x):
+        return x == 2
+    assert p2._error is "ERROR2" and p2._ignore_extra_keys
```
gst: you should use `==` to compare strings

PR #149, opened 2018-05-03
Fix callable check under PyPy2

cfs-pure:
With PyPy (at least version 2.7.3), methodcaller generated callables do have a __name__ attribute, so _callable_str() will not attempt to stringify it making this unit fail.


skorokithakis:
Thank you!

Issue #134, opened 2017-03-10
Collect all errors before throwing them


dAnjou:
Instead of raising an error as soon as possible it would be nice to collect all of them and return or raise them all at once.

sjortiz:
@keleshev @dAnjou  I would like to work on this

skorokithakis:
Please feel free. The only thing I'd like to see is making the error as user-friendly as possible, as I generally like to pass the error to the users when doing things like validating an API's input data.

mlhamel:
Just my two cents,

Maybe a first step that could be convenient just adding something:

```python
def is_valid(self, data):
    try:
        self.validate(data)
        return True
    except SchemaError:
        return False
```

Some time you just not care of what happen and you just want to know if it pass. Then there's the question of what should you do if you want to get the errors after the fact.

Collecting the errors and then returning them is not necessarily a bad idea. Although `validate` became huge enough that the coupling might be difficult to play with.

skorokithakis:
Hmm, yes, I can see why you don't like either option. Incidentally, I like @mlhamel's suggestion, as there were many times when I just wanted to know if something validates. If anyone would like to submit a small PR for that, I'd be grateful.

As for the problem at hand, we could go with the first approach and add that as an attribute to the exception that was raised, something like `e.errors` or similar, but yeah, I'm not a huge fan...

sjortiz:
I'll implement @mlhamel solution today/tomorrow.

The other one (the one in question), yeah we can do e.errors and see if I can add a new type of exception called "multi_errors" (or something alike) that it raises message consist in displaying the values in the array, after that I'll post a sample here, I'm not sure yet but, I think it won't have the traceback.

mlhamel:
Let me know how it goes, I would be happy to give hand also !

sjortiz:
Sorry for the delay on this guys.
@mlhamel that's the PR that includes your feature.
@skorokithakis Would be really good if you can review this, as I'm not able to assign reviewers.

sjortiz:
also I realized the fact that the current building doesn't support type hitting in any python version, might be worth checking/requesting in a new issue if you think it will be useful.
@skorokithakis

skorokithakis:
Sorry @sjortiz, which PR are you referring to?

dlobue:
I'd love to see this feature as well!

vinayak-mehta:
Are there any plans to implement this? Would love to get all validation errors.

skorokithakis:
[...]

PR #150, opened 2018-05-08
creating is_valid method as suggested in the issue 134

sjortiz:
Originated from the issue #134 

Description:
This methods returns True if the given data has passed all the validations that were specified in the given schema, returns false if not.

Considerations:

This method doesn't raise any exception and thus prevent any exception from the schema validation to pass silently

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -237,10 +237,8 @@ class Schema(object):
         """
         try:
             self.validate(data)
-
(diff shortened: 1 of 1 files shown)


skorokithakis:
Thank you!

mlhamel:
❤️

PR #153, opened 2018-06-10
Add "Constrained" container schema validator

gst:
Hi,

Made this as a POC for one my needs (replace a JSON validator that can handle array/lists with specific number of items and with specific type(s) for the items given their position).

Should work with any "container" type that can be indexed, and taking as in input a generator of the items to be collected.

Thanks to reviewers.
I hope this wasn't already doable with some other way ;)

Regards,


gst:
obviously that would require its own doc part in README but I'm waiting to see what reviewers might think over the idea & initial poc implementation (and tests :p)..

skorokithakis:
Thank you for your contribution! Can you explain a bit more about why this is useful? I'm not getting the exact use case from the code :/

gst:
it's to be able to define _array_ (I've in mind list or tuple but any container type accepting an iterable would be suitable) having a well defined structure.
maybe the tests should give you a better idea of the feature.. ?
but basically it's to be able to restrict the accepted / validated exact **structure** of container (with some extent possible: _additionals_ with min+max items).
maybe the name ConstrainedContainer isn't the best one too.. (I'm not native english).

PR #155, opened 2018-06-28
Added Py36 support

brunobord:
Tests look green on my local machine, I think it'd be a good idea to add it to the supported Python versions.


skorokithakis:
Thanks!

PR #159, opened 2018-08-21
Some performance improvements

some-0n3:
Hi,
I like your library, but found it a bit slow. I tried to speed up things, which made the code a little more messy. But in my tests it run roughly twice as fast.
I tested it with 10k examples from a wiktionay dump. The tests were mostly positive (all examples passed), but  contained some `Or` schemas that failed regularly.

I run the tests with python 2.7, 3.6 and  pypy, the pypy tests had a skip, all the others were successful.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -218,10 +218,8 @@ def _priority(s):
 
 
 def _flattable(schema):
-    """Return if a schema can be flattened."""
-    if isinstance(schema, Schema):
-        return type(schema) in (Schema, Optional, Forbidden, Const)
-    return isinstance(schema, Base)
+    """Return if the wrapping can be ommitted."""
+    return schema in (schemify, Schema, Forbidden, Optional)
 
 
 def _empty(schema):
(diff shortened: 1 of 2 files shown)


skorokithakis:
Thanks for this! It's a large PR, so it might take a bit of time to review.

some-0n3:
Ok, this is (hopefully) my last edit for this.
Sorry that the first commit was abit too messy, had some trouble with tox (my native python is python 3).
Now everything should be covered (except  "is_valid") and somewhat more readable.

skorokithakis:
@rmorshea It's going slowly because it's pretty large and I'm swamped these days :/ Have you reviewed it? If so, I'll trust your assessment (although the conflicts will need to be resolved).

rmorshea:
@skorokithakis I have not reviewed it, however I'll see if I can make time this week.

rmorshea:
@some-0n3 do you think you'd be able to hop on a call to chat about this. Maybe google hangouts? I think that the changes here are good. It's just hard to review this without a solid hour to do so. This is just going to keep getting pushed back and refactoring will become impossible.

If so, what time zone are you in. I'm PST.

PR #160, opened 2018-08-25
Remove ITERABLE type validating.

Exahilosys:
There's no need to automatically type check ITERABLE flavours when this can be done by the user. 
The reason this is needed is because different iterables can often be treated the same way.
For example, the default JSON encoder accepts both tuples and lists as array types.
Creating a schema with either of them will result in the validation failing if the other is passed instead.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -249,7 +249,6 @@ class Schema(object):
         i = self._ignore_extra_keys
         flavor = _priority(s)
         if flavor == ITERABLE:
-            data = Schema(type(s), error=e).validate(data)
             o = Or(*s, error=e, schema=Schema, ignore_extra_keys=i)
             return type(data)(o.validate(d) for d in data)
         if flavor == DICT:


skorokithakis:
Hmm, I'm not sure what you mean. Doesn't iterable accept both lists and tuples? I'm confused as to the reasoning behind the removal.

Exahilosys:
Consider the following data:
```py
data = tuple(range(1))
```
And the following schema:
```
schema = Schema([int])
```
Validating raises the following error:
```py
schema.SchemaUnexpectedTypeError: (0,) should be instance of 'list'
```
As `.validate` automatically validates the iterable's type.
Removing that line of code passes the decision of type validation to the user.
By adding a bit of extra logic in script code, the same results can be achieved:
```
schema = Schema(And(list, Schema([int])))
```
All while allowing to further abstractify or even bypass such validation:
```
import collections.abc
schema = Schema(And(collections.abc.Iterable, Schema([int]))) # abstract
schema = Schema([int]) # bypass
```
Using `list` to create the data and `(int,)` in schema would have the the same error failing for `tuple`.


This request arose when I tried to validate either `tuple` or `list` (without knowing what to expect) against an `ITERABLE` flavoured schema while working with `JSON` since the default encoder accepts both.

PR #169, opened 2018-10-02
Schema Composition (Do Not Merge)

### LINKED ISSUE #168: Ability to Compose Schemas Together
It would be great if it were possible to compose two schemas together:

```python
s1 = Schema({"a": {"b": int}})
s2 = Schema({"a": {"b": int}})

s1_and_s2 = compose(s1, s2)

assert s1_and_s2 == Schema({"a": {"b": int, "c": int}})
```

If such a proposal is reasonable and possible I am willing to create a PR. Suggestions are welcome!

rmorshea:
# Summary

Closes #168 

With the `compose` function two or more schemas can be merged into one. The composition of schema types is handled in three cases:

1. All are dictionaries.
   * The schemas are merged together into one dictionary schema where shared keys are recursively composed.
2. All are iterables.
   * All schemas are merged into a single iterable schema in the order they were passed into ``compose``.
3. There is a combination of different schema types.
   * A type error is raised if dictionary and iterable type schemas are composed.
   * Any other combination of schema types are joined using a ``function(*schemas)`` (defaults to ``And``).

# Examples

Compose two schemas together using the default reducer.

```python
s1 = Schema({"a": int, "b": int})
s2 = Schema({"b": float})
s3 = compose(s2, s1) # note choice of order
assert s3.is_valid({"a": 1, "b":, 2.0})
```

Compose two schemas using a "validator" reducer where the resulting schema requires all conditions from the given schemas to be true. In this particular case the merged schema validates when given a string that is lowercase.

```python
s1 = Schema(str)
s2 = Schema(lambda s: s.lower() == s)
and_reducer = Use(lambda schemas: And(*schemas))
s3 = compose(s1, s2, reduce=and_reducer)
assert s3.is_valid("hello world!")
assert not s3.is_valid("Hello World!")
```

The following example demonstrates how to form more complex reducers by using an `Or` validator to combine to combine others (including the `and_reducer`) from the example above.

```python
import functools, operator
s1 = Schema({"a": [int], "b": str})
s2 = Schema({"a": [float], "b": lambda s: s.lower() == s})
join_lists = lambda schemas: functools.reduce(operator.add, schemas, [])
list_reducer = And([list], Use(join_lists))
reducer = Or(list_reducer, and_reducer)
s3 = compose(s1, s2, reduce=reducer)
assert s3.is_valid({"a": [1, 2.0], "b": "hello world"})
```

# Gist

https://gist.github.com/rmorshea/1e8d00dbd9f583f763837c47d3910c25

# Todo

+ Update README
+ Add tests.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -51,7 +51,7 @@ def compose(*schemas, **kwargs):
 
     Notes
     -----
-    The composition of schema types is handled in four cases:
+    The composition of schema types is handled in three cases:
 
     1. All are dictionaries.
 
@@ -61,10 +61,10 @@ def compose(*schemas, **kwargs):
 
        * All schemas are merged into a single iterable schema in the order they were passed into ``compose``.
 
-    4. There is a combination of different schema types.
+    3. There is a combination of different schema types.
 
        * A type error is raised if dictionary and iterable type schemas are composed.
-       * Any other combination of schema types are merged under one ``And`` validator.
+       * Any other combination of schema types are joined using a ``function(*schemas)`` (defaults to ``And``).
     """
     join = kwargs.get("join", And)
     schemas = [s._schema if isinstance(s, Schema) else s for s in schemas]
--- a/schema.py
+++ b/schema.py
@@ -21,78 +21,80 @@ def compose(*schemas, **kwargs):
     ----------
     *schemas: any
         A series of values to be interpreted as Schemas.
-    join: callable (default: ``And``)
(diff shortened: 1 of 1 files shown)


skorokithakis:
I like the idea in this, and it helps that it solves #171. Will review soon, thank you!

rmorshea:
@skorokithakis don't merge yet! I need to write tests.

rmorshea:
Closing for now. I don't expect I'll get to this one.

PR #170, opened 2018-10-03
Remove Assert Statements

rmorshea:
[`assert` statements are bypassed](https://stackoverflow.com/questions/1273211/disable-assertions-in-python) when running in "optimized" mode (i.e. `python -O`).

`assert` should only be used in tests or for debug output.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -73,7 +73,9 @@ class And(object):
     """
     def __init__(self, *args, **kw):
         self._args = args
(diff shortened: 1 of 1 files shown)


skorokithakis:
I agree with this, but it's backwards-incompatible because it'll break people who expect an `AssertionError`. Regardless, I will merge it because it's grossly inappropriate for us to be using `assert`s t here.

skorokithakis:
Thank you!

PR #174, opened 2018-10-17
Allow exclusive or in a dict schema

### LINKED ISSUE #106: XOr implementation
First of all, this library is awesome.

I've implemented a `XOr` class for exclusive schemas (shamelessly copying from the `Or` class). Maybe this is naively wrong, but so far this has worked as expected.

``` python
class XOr(And):

    def validate(self, data):
        accepted = False
        x = SchemaError([], [])
        for s in [Schema(s, error=self._error) for s in self._args]:
            try:
                validated = s.validate(data)
                if accepted:
                    raise SchemaError(['%r did not validate %r' % (self, data)],
                                      [self._error])
                accepted = True
            except SchemaError as _x:
                x = _x
        if not accepted:
            raise SchemaError(['%r did not validate %r' % (self, data)] + x.autos,
                              [self._error] + x.errors)
        return validated
```

If you are interested in this, I can open a PR.

julienduchesne:
Fixes #106 

It's a bit sad that `Schema` has to know that `Or` has a `reset()` method but `Schema` has to know when to reset. Otherwise, subsequent `validate` calls of the same schema with the same value will fail. 
The other thing I considered is to subclass all the validators with a base class with an empty `reset()` method and always call it, `Or` would have overridden that method.

Let me know what you think!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -268,6 +268,20 @@ data matches:
 Defaults are used verbatim, not passed through any validators specified in the
 value.
 
+In a dictionary, you can also combine two keys in a "one or the other" manner. To do
+so, use the `Or` class as a key:
+
+.. code:: python
+    from schema import Or, Schema
+    schema = Schema({
+        Or("key1", "key2", only_one=True): str
+    })
+
+    schema.validate({"key1": "test"}) # Ok
+    schema.validate({"key1": "test", "key2": "test"}) # SchemaWrongKeyError
+
+    
+
 You can mark a key as forbidden as follows:
 
 .. code:: python
--- a/schema.py
+++ b/schema.py
@@ -71,6 +71,7 @@ class And(object):
     """
     Utility function to combine validation directives in AND Boolean fashion.
     """
+
     def __init__(self, *args, **kw):
         self._args = args
         if not set(kw).issubset({'error', 'schema', 'ignore_extra_keys'}):
@@ -102,6 +103,15 @@ class And(object):
 class Or(And):
     """Utility function to combine validation directives in a OR Boolean
     fashion."""
+
+    def __init__(self, *args, **kwargs):
+        self.only_one = kwargs.pop('only_one', False)
+        self.reset()
+        super(Or, self).__init__(*args, **kwargs)
+
+    def reset(self):
+        self.got_one = False
+
     def validate(self, data):
         """
         Validate data using sub defined schema/expressions ensuring at least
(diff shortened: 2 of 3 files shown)


skorokithakis:
Sorry, can you post a bit of code that illustrates the problem? I'm not sure I got it from the description...

julienduchesne:
Oh sorry, it's mostly from #106. In short, Or("key", "newkey") in a dict will accept 
```{
  "key": value",
  "newkey": "value"
}```
This PR allows for an exclusive Or. This is useful in a deprecation use case. You want the user to switch to the new key name but you still want to support the old one. However, setting both can cause issues so it should be rejected.

julienduchesne:
Oh sorry, it's mostly from #106. In short, Or("key", "newkey") in a dict will accept 
```{
  "key": value",
  "newkey": "value"
}```
This PR allows for an exclusive Or. This is useful in a deprecation use case. You want the user to switch to the new key name but you still want to support the old one. However, setting both can cause issues so it should be rejected.

julienduchesne:
Oh sorry, it's mostly from #106. In short, Or("key", "newkey") in a dict will accept 
```{
  "key": value",
  "newkey": "value"
}```
This PR allows for an exclusive Or. This is useful in a deprecation use case. You want the user to switch to the new key name but you still want to support the old one. However, setting both can cause issues so it should be rejected.

julienduchesne:
Oh sorry, it's mostly from #106. In short, Or("key", "newkey") in a dict will accept 
```{
  "key": value",
  "newkey": "value"
}```
This PR allows for an exclusive Or. This is useful in a deprecation use case. You want the user to switch to the new key name but you still want to support the old one. However, setting both can cause issues so it should be rejected.

julienduchesne:
Oh sorry, it's mostly from #106. In short, Or("key", "newkey") in a dict will accept 
```{
  "key": value",
  "newkey": "value"
}```
This PR allows for an exclusive Or. This is useful in a deprecation use case. You want the user to switch to the new key name but you still want to support the old one. However, setting both can cause issues so it should be rejected.

skorokithakis:
Oh sorry, I see what you mean. I thought your comment was the motivation for this PR, but I now realize it's a comment on the implementation. From a first look, this looks good, I will review in more depth soon!

skorokithakis:
Github is having some trouble, I saw your comment posted ten times, I tried to delete a few but they went away again. I'll return later when Github is feeling better.

skorokithakis:
Also, can you remove the [WIP] if this is good to merge?

julienduchesne:
Github did not trigger travis but it should be ready now

skorokithakis:
Hmm, I can't access the online conflict resolution UI either... I just merged your other PR, but it conflicts with this one, can you resolve so I can merge this too?

PR #175, opened 2018-10-18
Add Hook class. Allows to introduce custom handlers

julienduchesne:
* Add Hook class that allows custom handlers
* Make Forbidden a subclass of Hook that raises an exception as its handler
* My personal use case for this, creating a Deprecated hook that will be called that way: `Deprecated("attribute")`. My handler will simply be a function that logs a warning to the user when he uses that attribute. I am sure there are other use cases for a custom handler class.

Let me know what you think!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -356,7 +356,7 @@ Here's an example where a `Deprecated` class is added to log warnings whenever a
 
     Schema({Deprecated("test", "custom error message."): object}, ignore_extra_keys=True).validate({"test": "value"})
     ...
-    >>> WARNING: `test` is deprecated. custom error message.
+    WARNING: `test` is deprecated. custom error message.
 
 Extra Keys
 ~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


julienduchesne:
For further info on my use case, I simply wrote this snippet. 
```
class Deprecated(Hook):
    def __init__(self, *args, **kwargs):
        kwargs["handler"] = lambda key, *args: logging.warn(f"`{key}` is deprecated. " + (self._error or ""))
        super(Deprecated, self).__init__(*args, **kwargs)
```
And I use it it this way: `Deprecated("service", "Use `product` instead."): object,`

skorokithakis:
Thanks for the PR! I like this functionality, but I think we need a bit of documentation about what it does and how people can use it for the PR to be complete.

julienduchesne:
Added some docs

skorokithakis:
This looks good, thanks! The README fails to compile, I think there's a syntax error in the RST, but I'm good to merge once that's fixed.

julienduchesne:
github did not trigger travis but it should be fixed now.

skorokithakis:
Hmm, I wonder why Travis isn't triggering... Thanks!

PR #178, opened 2018-10-24
Fix Or reset() when Or is Optional

julienduchesne:
This fixes an issue in https://github.com/keleshev/schema/pull/174. When Or was Optional, the reset method was not called

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -283,10 +283,12 @@ class Schema(object):
             # for each key and value find a schema entry matching them, if any
             sorted_skeys = sorted(s, key=self._dict_key_priority)
             for skey in sorted_skeys:
-                if getattr(skey, "reset", None):
+                if hasattr(skey, "reset"):
                     skey.reset()
 
-            for key, value in data.items():
+            # Evaluate dictionaries last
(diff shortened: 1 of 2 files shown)


skorokithakis:
This looks good to me, is it good to merge?

julienduchesne:
Well tests fail but I'm a bit puzzled. Tox runs fine on my machine :tm:

skorokithakis:
Yeah, that's very odd, and it works on 3.6, and I don't see any backwards-incompatible features used. I've restarted the job, we'll see.

julienduchesne:
I managed to make it fail locally. It is very flaky, weirdly. I will investigate and comment again when it is fixed

skorokithakis:
No, it fails, this is extremely odd. Does it work for you under 3.5?

skorokithakis:
Ah, alright, thanks!

skorokithakis:
It was the guaranteed dictionary order, wasn't it.

julienduchesne:
Yes, it only seemed to fail in 3.5. At least on my machine. Now it's pep8 :man_facepalming:

skorokithakis:
Thank you!

PR #179, opened 2018-10-29
Travis CI python 3.7

cxong:
Note:
- Travis CI requires `dist: xenial` + `sudo: true` for python 3.7 (https://github.com/travis-ci/travis-ci/issues/9815). These statements can be removed once they move to xenial by default
- For some reason my pep8 build is using pyflakes 2.0.0 which is performing W504 check by default, which would cause errors in `schema.py`. I suppressed that warning in tox.ini instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,9 @@ matrix:
     - env: TOXENV=py37
       python: 3.7
       dist: xenial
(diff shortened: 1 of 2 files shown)


skorokithakis:
Looks good, thank you!

PR #180, opened 2018-10-29
Generate JSON schema from schema

julienduchesne:
Big-ish PR but the title is self explanatory. Implementation of draft-7 json schema generation: https://json-schema.org/.
Some caveats:
- Some types are not handled. In that case, the `{}` dict is returned (no checks in that case)
- There are no object references, objects are always fully rendered
- Also there may be other features of JSON Schema that I did not implement (because I'm not aware of them :smile: )
- I tested it with a very large schema, and all of it works. I'd love to see some use cases that I did not think of though
- I did not do any documentation, if we do any, it should probably be noted that this is a beta feature. And I wanted to have your opinion on this feature before writing any of it.

Let me know what you think!

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -3,7 +3,6 @@ obtained from config-files, forms, external services or command-line
 parsing, converted from JSON/YAML (or something else) to Python data-types."""
 
 import re
-import json
 
 __version__ = '0.6.8'
 __all__ = ['Schema',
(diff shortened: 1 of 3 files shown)


skorokithakis:
Hmm, this looks interesting, but I don't understand the reason for this to be in Schema. Can you elaborate on why you think people would find it useful?

schema.py
```diff
@@ -379,6 +379,77 @@ def validate(self, data):
             raise SchemaError('%r does not match %r' % (s, data),
                               e.format(data) if e else None)
 
+    def json_schema(self, schema_id=None, is_main_schema=True):
```
skorokithakis: Can you add a docstring here to briefly explain what this method does?

skorokithakis:
I agree with you about the documentation pointing out that this is in beta, but I love this feature, and would like to see it merged. As it stands, as long as we clearly state in the docs that it's not bulletproof, I'm good to merge.

skorokithakis:
Looks great, thank you @julienduchesne!

Issue #143, opened 2017-08-11
Validation depending on attribute of the instance


CausalReinforcer:
Is there any chance to enable the following function:

In the schema, there is an 'pet_type', which may take value from ('dog', 'cat'). There is also an attribute of 'pet_food', whose legitimate value depends on 'pet_type' - if it is a dog, then 'pet_food' can take value from ('dog_food_1', 'dog_food_2') , and if it is a cat, then 'pet_food' can take value from ('cat_food_1', 'cat_food_2').

This is quite needed for my ongoing project. Thank you!

towr:
I'm looking for a similar feature. in my case that given two dates one is smaller than another.
Can items really only be validated in isolation?

skorokithakis:
So far, I'm afraid so. I'm not sure what it would take to change this, as the library is written with fields being independent of each other.

towr:
I've just realized that in my case I can get what I want with something like 
`And({'from':str, 'to':str}, lambda d:d['from'] < d['to']).validate({'from':'2016-01-01', 'to':'2017-01-01'})`
Since I'm validating a dict, I can also use a validation function on the dict as a whole.

PR #181, opened 2018-11-06
When ignoring extra keys,  Or's only_one should still be handled

julienduchesne:
Sorry for another PR about this. I noticed that Or's `only_one` condition didn't work when mixed with `ìgnore_extra_keys` as it was relying on the `WrongKey` exception. I instead implemented a type of Error that stops the execution immediately.

Let me know if you see anything I might have missed. 
Thanks

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -15,9 +15,11 @@ __all__ = ['Schema',
 
 
 class SchemaError(Exception):
-    """Error during Schema validation."""
+    """Error during Schema validation.
+    If the Error should stop the execution immediately, set should_stop to True"""
 
-    def __init__(self, autos, errors=None):
+    def __init__(self, autos, errors=None, should_stop=False):
+        self.should_stop = should_stop
         self.autos = autos if type(autos) is list else [autos]
         self.errors = errors if type(errors) is list else [errors]
         Exception.__init__(self, self.code)
@@ -53,6 +55,12 @@ class SchemaMissingKeyError(SchemaError):
     """Error should be raised when a mandatory key is not found within the
     data set being validated"""
     pass
+
+
+class SchemaOnlyOneAllowedError(SchemaError):
+    """Error should be raised when an only_one Or key has multiple matching candidates"""
+    def __init__(self, autos, errors=None):
+        SchemaError.__init__(self, autos, errors, should_stop=True)
 
 
 class SchemaForbiddenKeyError(SchemaError):
@@ -110,7 +118,7 @@ class Or(And):
         super(Or, self).__init__(*args, **kwargs)
 
     def reset(self):
-        self.got_one = False
+        self.match_count = 0
 
     def validate(self, data):
         """
@@ -125,12 +133,15 @@ class Or(And):
                   for s in self._args]:
             try:
                 validation = s.validate(data)
-                if self.got_one and self.only_one:
+                self.match_count += 1
+                if self.match_count > 1 and self.only_one:
                     break
-                self.got_one = True
                 return validation
             except SchemaError as _x:
                 autos, errors = _x.autos, _x.errors
+        if self.match_count > 1 and self.only_one:
+            raise SchemaOnlyOneAllowedError(['There are multiple keys present ' +
+                                             'from the %r condition' % self])
         raise SchemaError(['%r did not validate %r' % (self, data)] + autos,
                           [self._error.format(data) if self._error else None] +
                           errors)
@@ -294,8 +305,9 @@ class Schema(object):
                     svalue = s[skey]
                     try:
                         nkey = Schema(skey, error=e).validate(key)
-                    except SchemaError:
-                        pass
+                    except SchemaError as x:
+                        if x.should_stop:
+                            raise x
                     else:
                         if isinstance(skey, Hook):
                             # As the content of the value makes little sense for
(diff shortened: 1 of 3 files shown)


skorokithakis:
This looks good to me. @rmorshea do you want to take a look too?

rmorshea:
@julienduchesne the purpose of `should_stop` isn't obvious to me. Could you explain why its needed? Maybe this explanation should be documented in addition to what you've already have? Outside of this it looks good to me.

@skorokithakis this may be something that should be sorted out in a separate PR, but it seems like `Or.reset()` should be called at the end of `Or.validate` instead of relying on it being called [here](https://github.com/keleshev/schema/pull/181/files#diff-f9bdc6e9ac5d7d6096bf607cc6260f9eR297).

julienduchesne:
Yes, I struggled to find a good name for this attribute. What happens is that SchemaErrors are caught and ignored (meaning that a validator has failed). However, in the case of Or's only_one, we want to stop straight away because if the exception is caught, it would simply mean that key was not matched when in fact it failed the validator. I'm not sure if that's clear.

For the Or.reset(), it has to be outside the Or.validate() because it relies on outside context. Ex: You have `key1` and `key2`. `Validate()` will be called once for `key1` and once for `key2`, however it has to succeed at `key1` but fail at `key2`. Only the Schema class that contains the `Or` condition knows when to call reset(). It seemed to be the issue here: https://github.com/keleshev/schema/issues/106

rmorshea:
@julienduchesne I see now.

What if `reset` raised the `SchemaOnlyOneAllowedError`? You could accomplish this using [`contextlib.ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) (or the Python 2 equivalent). Consider this:

```python
from contextlib import ExitStack

exitstack = ExitStack()

for k in schema:
    if hasattr(k, "reset"):
        exitstack.callback(k.reset)

with exitstack as es:
    # validate the keys...
```

With this design you solve the problem of having an intuitive way to manage the `reset` of keys, and the problem of raising `SchemaOnlyOneAllowedError` without needing `should_stop`.

It's possible that you could extend this context manager approach to enable other features. Maybe https://github.com/keleshev/schema/issues/143?

julienduchesne:
I like this suggestion! :+1:. I made the changes, though I had a bit of trouble importing contextlib as setup.py imports schema so importing something in schema would fail as there cannot be any external dependencies when running the setup.py in tox. I also setup tox to recreate envs automatically because it is safer when there are dependencies. 

1. Would you like the recreate better in the .travis.yml config instead (Adding `-r`). 
2. Do you have any suggestions on how to handle the deps other than the try except?

rmorshea:
@julienduchesne so I think the correct way to implement the try-except for the import would be:

```python
try:
    from contextlib import ExitStack
except ImportError:
    from contextlib2 import ExitStack
```

Also is there a reason that `exitstack` is a global variable? I can see that causing some very strange bugs. I was expecting to see that variable assignment inside of `validate`.

rmorshea:
@julienduchesne I don't really know travis that well, so I can't speak to `1.` - maybe @skorokithakis can?

rmorshea:
As for `setup.py` we probably shouldn't be importing the very package we want to install/build. So I think we can hard code the package name to `schema`. As for `__version__`, In the past I've done this:

```python
project_root = "path/to/project_root.py"
with open(project_root) as f:
    for line in f.read().split("\n"):
        if line.startswith("__version__ ="):
            version = eval(line.split("=", 1)[1])
    else:
        path = os.path.join(root, '__init__.py')
        print("No __version__ attribute found in %r" % project_root)
        sys.exit(1)
```

Or create a `_version.py` and `exec` it:

```python
# in _version.py
__version__ = "0.1.0"
```

```python
# in setup.py
version_module = "path/to/_version.py"
with open(version_module) as f:
    namespace = {}
    exec(f.read(), {}, namespace)
    version = namespace["__version__"]
```

I think I prefer the first solution since it doesn't require the extra `_version.py` file. It's kind of up to @skorokithakis though.

schema.py
```diff
@@ -4,14 +4,24 @@
 
 import re
 
+# The setup.py (which installs dependencies) needs access to the __version__
+# and __all__ attributes. Therefore, external dependencies and their global
+# usage need to be wrapped in a try/except.
+try:
+    from contextlib2 import ExitStack
+    exitstack = ExitStack()
```
rmorshea: I don't think the `exitstack` should be global. Best to assign it in `validate`.

skorokithakis:
Hey guys, sorry for the delay, just catching up. I agree with everything @rmorshea said, for the version handling I usually put the version in the `__init__.py` and parse that, like @rmorshea said.

Also, I agree with recreating the environments, as it's better if things run without any previous code.

Also, @rmorshea, can I interest you in joining a project I'm working on now? It's called [Code Shelter](https://www.codeshelter.co/faq/) and it's about volunteering to help out with unmaintained project.

rmorshea:
I think this looks good to me.

rmorshea:
@skorokithakis what are your expectations around time commitments for Code Shelter?

skorokithakis:
@rmorshea There are none, you put in as much time as you want/can.

rmorshea:
@skorokithakis haha well I think I can commit to that. Right now there's only one project. Are there others that you think would benefit from this?

skorokithakis:
Yes, you will receive an email every time a project is added. Right now I'm looking for maintainers, and the next step is to look for projects! If you click "maintainer login", I will approve your account.

rmorshea:
@skorokithakis done!

skorokithakis:
Approved, thank you!

PR #182, opened 2018-11-06
Custom Schema Names

SnapperGee:
In Schema.py

Added parameter "name" to the Schema class that defaults to an empty string.

Added the function set_schema_name() to the Schema class that formats and returns the Schema name if it isn't empty. set_schema_name() also takes in one string argument so the way it formats the name can be altered based on the schema error type.

The formatted name it returns is then used to be substituted into the error message that gets printed when a schema error is raised.


schema.py
```diff
@@ -233,10 +233,11 @@ class Schema(object):
     schema for the data that will be validated.
     """
 
-    def __init__(self, schema, error=None, ignore_extra_keys=False):
+    def __init__(self, schema, error=None, ignore_extra_keys=False, name=''):
```
rmorshea: `name` should probably default to `None` since it makes it more obvious that the parameter was not set.

```suggestion
    def __init__(self, schema, error=None, ignore_extra_keys=False, name=None):
```

schema.py
```diff
@@ -266,6 +267,26 @@ def is_valid(self, data):
         else:
             return True
 
+
+    def set_schema_name(self, error_type):
```
rmorshea: This should be a "private" method (i.e. prefixed with an underscore `_`).

Also, I might suggest changing this to be `_raise_error(self, exception, message)` which prepends `self._name` to message if it's not `None` then raises `exception(self._name + " " + message)`. Example usage:

```python
message = "missing keys {keys}".format(keys=the_missing_keys)
self._raise_error(SchemaMissingKeyError, message)
```

rmorshea: If it's possible to make this method the one stop shop for error raising inside `validate` that would be amazing since it would clean up a log of the code dedicated to message formatting in it.

schema.py
SnapperGee: That's how I actually originally had it implemented for the very reason you suggested, but ultimately figured not needing the additional conversion from `None` to `' '` would be beneficial. But having the added explicitly of converting it certainly wouldn't hurt and is simple enough to implement.

schema.py
SnapperGee: Big +1 on this
This is ultimately what I'm hoping for and what I originally attempted until I realized accessing the various variables that are unique to each schema error slightly complicates things. Not terribly, but enough that I didn't wanna try to implement anything too drastic right off the bat while I was still getting a feel for the Schema module. But ya, I definitely agree

schema.py
skorokithakis: I agree with @rmorshea here, we don't need this right now but it's wrong semantically and in the future we might need a distinction between the empty string and `None`, so I wouldn't want to make the API wrong here now.

schema.py
skorokithakis: Agreed as well.

schema.py
```diff
@@ -266,6 +267,26 @@ def is_valid(self, data):
...
+        """Returns string to add to error message based on what the error type
+        is and whether a custom schema name has been defined or not.
+        """
+        if self._name != '':
```
skorokithakis: This can just be:

```
if not self._name:
    return ""
```

And you save yourself a whole indentation level.

schema.py
```diff
@@ -266,6 +267,26 @@ def is_valid(self, data):
...
+            if error_type == "SchemaError":
+                error_message = "\'%s\' " % (self._name,)
```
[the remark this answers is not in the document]
skorokithakis: I think @rmorshea is referring to the `%r`, which you can use to get a representation of the string, so you don't have to add the quotes yourself. Basically, this whole section can be:

```
if error_type in ("SchemaError", "MissingKey", "WrongKey", "UnexpectedType":
    return repr(self._name)
```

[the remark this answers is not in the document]
SnapperGee: Right now, since all the schema names are formatted the same way for each error type
`"\'%s\' " % (self._name,)`
it is redundant. But I was thinking the additional flexibility of being able to make the schema names' formatting unique to the error type it's appended to could be beneficial since some of the schema error messages that get raised can vary a bit depending on the the type of schema error it is

SnapperGee:
Closed for now till all changes get successfully implemented. Don't need ya guys badgered with changes till it's more complete

schema.py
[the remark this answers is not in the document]
skorokithakis: Sure, but this is easy to change later, so YAGNI dictates that we should go with the simplest implementation for now.

skorokithakis:
That's fine, you don't need to close it, it should stay open as it's active.

schema.py
```diff
@@ -267,22 +267,17 @@ def is_valid(self, data):
         else:
             return True
 
-    def _raise_error(self, error_type):
-        """Returns string to add to error message based on what the error type
-        is and whether a custom schema name has been defined or not.
+    def _raise_error(self, exception, message, *error_list):
```
rmorshea: Is there a case where there is more than one error passed to `error_list`? Right now you only process `error_list[0]`. No need to lock yourself into a design decision (i.e. using up `*args`) if you don't have to.

Also, what if you simply accessed `self._error` in this method instead of passing it in? That would eliminate the need for [`format_error`](https://github.com/keleshev/schema/pull/182/files#diff-f9bdc6e9ac5d7d6096bf607cc6260f9eR285) in `validate` since you could simply house that logic inside this method.

schema.py
```diff
@@ -266,22 +267,34 @@ def is_valid(self, data):
...
+            raise exception("%r %s" % (self._name, message,),
+                            error_list[0])
+
     def validate(self, data):
         Schema = self.__class__
-        s = self._schema
-        e = self._error
         i = self._ignore_extra_keys
-        flavor = _priority(s)
+        def format_error(e): return e.format(data) if e else None
```
rmorshea: I think this logic can be a part of `_raise_error`.

schema.py
```diff
@@ -266,22 +267,34 @@ def is_valid(self, data):
...
+        flavor = _priority(self._schema)
         if flavor == ITERABLE:
-            data = Schema(type(s), error=e).validate(data)
-            o = Or(*s, error=e, schema=Schema, ignore_extra_keys=i)
+            data = Schema(type(self._schema), error=self._error).validate(data)
+            s = self._schema
```
rmorshea: No need for the assignment here.

schema.py
SnapperGee: This flexibility is necessary since there's at least one instances where the `*error_list` argument isn't given a value so it's necessary to make this parameter optional and when the `*error_list` parameter is defaulted to `None`, validations will pass when they should fail.

On line 389:
`self._raise_error(SchemaError, message)`
No 3rd argument given.

I looked into including the `format_error()` function into `_raise_error()`. Unfortunately there are some instances where `self._error` is accessed directly without being formatted (via `self._error.format(data)`) or set to `None` and other instances where the `*error_list` argument is given a value where `self._error` is neither directly accessed nor formatted.

One line 380 to 381:
`self._raise_error(SchemaError, message,
                                  format_error(self._error))`
A `SchemaError` is raised that formats or sets `self._error` to `None`

On line 395:
`self._raise_error(SchemaError, message, self._error)`
Another `SchemaError` is raised, except `self._error` is accessed directly.

There are also other instances where various schema errors access `self._error` both directly and after formatting it.

On line 376-377:
`self._raise_error(SchemaError, [None] + x.autos,
                                  [self._error] + x.errors)`
`self._error` is neither formatted, set to none, nor just simply accessed directly.

So for the above reasons I didn't want `_raise_error()` to perform `format_error()` on the argument given for `*error_list` every time. And since `format_error()` is both applied and not applied for the same type of schema errors, I can't easily set `format_error()` to only be executed based on the exception that's raised.

The only solution I can think of at the moment is to have `self._raise_error()` take in a fourth argument that "flags" whether `format_error()` should be executed on the `*error_list` argument.

It's at this point that I figured, for now at least, it's simpler to leave `format_error()` external to `self._reaise_error()`

schema.py
SnapperGee: Taken from my above answer:

I looked into including the `format_error()` function into `_raise_error()`. Unfortunately there are some instances where `self._error` is accessed directly without being formatted (via `self._error.format(data)`) or set to `None` and other instances where the `*error_list` argument is given a value where `self._error` is neither directly accessed nor formatted.

One line 380 to 381:
`self._raise_error(SchemaError, message,
                                  format_error(self._error))`
A `SchemaError` is raised that formats or sets `self._error` to `None`

On line 395:
`self._raise_error(SchemaError, message, self._error)`
Another `SchemaError` is raised, except `self._error` is accessed directly.

There are also other instances where various schema errors access `self._error` both directly and after formatting it as well.

On line 376-377:
`self._raise_error(SchemaError, [None] + x.autos,
                                  [self._error] + x.errors)`
`self._error` is neither formatted, set to none, nor just simply accessed directly.

So for the above reasons I didn't want `_raise_error()` to perform `format_error()` on the argument given for `*error_list` every time. And since `format_error()` is both applied and not applied for the same type of schema errors, I can't easily set `format_error()` to only be executed based on the exception that's raised.

The only solution I can think of at the moment is to have `self._raise_error()` take in a fourth argument that "flags" whether `format_error()` should be executed on the `*error_list` argument.

It's at this point that I figured, for now at least, it's simpler to leave `format_error()` external to `self._reaise_error()`

schema.py
SnapperGee: The reason for this assignment is for line 290:
`o = Or(*`__s__`, error=self._error, schema=Schema, ignore_extra_keys=i)`

`s` is used in this assignment expression

rmorshea: @SnapperGee why not do `*self._schema` instead of put the assignment on the previous line.

schema.py
```diff
@@ -266,22 +267,34 @@ def is_valid(self, data):
         else:
             return True
 
+    def _raise_error(self, exception, message, *error_list):
+        """Raises exception with custom schema name prepended to exception
+        message and feeds additional arguments into schema error class.
+        """
+        if not self._name and not error_list:
+            raise exception("%s" % (message,))
```
rmorshea: just do `str(message)`

skorokithakis:
By the way, @rmorshea, would you be interested in becoming a maintainer for Schema? I can't add you, but I can maybe suggest it to @keleshev?

rmorshea:
I'd be ok with that. I don't think I'd be comfortable doing releases or anything like that, but I'm happy to help out with PRs and Issues when I have time.

skorokithakis:
That sounds good to me, thanks! I'll see if @keleshev can add you as a collaborator.

schema.py
SnapperGee: Default to `None` has been implemented

schema.py
SnapperGee: `_raise_error` method is currently implemented with the signature `(self, exception, message, error_list)` and generates the error message via `"%r %s" % (self._name, message,)` as well as raises the various errors within the function.

The function is always executed with a message so there's no need to have it default to None.

schema.py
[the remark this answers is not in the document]
SnapperGee: Now the way `self._raise_error()` operates, by taking in the "error_type" (now renamed to "exception"), "message", and "error_list", it makes checking the "error_type" redundant making the function itself simpler and shorter 👍.

Also `self._name` is now formatted with `repr` instead of `str`

skorokithakis:
@SnapperGee @rmorshea Is this good to merge (minus conflicts)?

SnapperGee:
> @SnapperGee @rmorshea Is this good to merge (minus conflicts)?

No, it's not good to merge yet. Gotta do some error debugging for certain instances. I'll let ya know though!

SnapperGee:
So far everything is looking pretty solid except for one issue. So far these changes only apply to the Schema class, so whenever the `And` class or `Or` class is invoked when there's an error, the custom schema name isn't applied to the error message that gets raised

schema.py
```diff
@@ -266,22 +267,61 @@ def is_valid(self, data):
...
+        else:
+            raised_exception = None
+
+        format_error = lambda e: e.format(data) if e else None
+        formatted_error = format_error(self._error)
+
+        formatted_schema_error_message = "{0!r} {1}".format(self._name, message)
+
+        if isinstance(message, list):
+            formatted_message = ''
```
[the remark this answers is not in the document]
SnapperGee: For loop, that concatenates message if message is a list, has been replaced with
```
if isinstance(message, list):
    message = " ".join(list(map(str, message)))
```

schema.py
```diff
@@ -266,22 +267,61 @@ def is_valid(self, data):
         else:
             return True
 
+    def _raise_error(self, schema_error, message, error_list, data=None,
+                     exception=None):
```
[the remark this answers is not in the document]
SnapperGee: It's their to make sure that
```
formatted_error = self._error.format(data) if self._error else None
```
is used correctly. I'm still in the process of making sure it's implemented and utilized correctly and will resolve this conversation once it's good to go

schema.py
```diff
@@ -266,22 +267,61 @@ def is_valid(self, data):
         else:
             return True
 
+    def _raise_error(self, schema_error, message, error_list, data=None,
```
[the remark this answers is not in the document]
SnapperGee: I think there are some instances where the `error_list` is `[self._error] + x.errors` and other times where it's just `self._error`. Having it prepended probably won't be a problem, but I just wanna make sure it doesn't break anything or is done in such a way that it causes problems.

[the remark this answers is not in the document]
rmorshea: If `error_list=None` is the default you could do:

```python
error_list = error_list or []
error_list.insert(0, self._error)
```

schema.py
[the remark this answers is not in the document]
SnapperGee: Is there a reason map has to be used here? Wouldn't `message = " ".join(message)` work since it's only executed if `message` is a list?

schema.py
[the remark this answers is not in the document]
SnapperGee: On line 382 and 393, there's
```
raise SchemaError([None] + x.autos, [e] + x.errors)
```
So that's behavior that'd have to be accounted for and implemented. For now, `[None] + x.autos` is being passed as the message argument and `[e] + x.errors` is being passed as the error_list argument to `raise_error` instead of `self._error` during those two instances which is why it's currently set up so `self._error` isn't hard coded into the function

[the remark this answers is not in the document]
SnapperGee: Made it so `self._error` is hard coded into `raise_error()` function. However, as a result, `None` had to be passed to `raise_error()` as `message` argument on line 412 and 422

SnapperGee:
Minus the conflicts, this pr should be good to merge. I updated the master branch so now it uses a much simpler function that only prepends the custom schema name if there is one.

skorokithakis:
Fantastic, but can you merge master into your branch and resolve the conflicts so I can merge?
 @rmorshea, does this look good to you?

SnapperGee:
I just rebased and pushed the changes. I did a little testing which it passed and will do more testing later or tomorrow unless you guys' think it looks good and it gets merged before then👍

SnapperGee:
I did some more testing and so far it's looking good

rmorshea:
@SnapperGee try pushing up a white-space change. It looks like one of the tests failed randomly.

Also can you add a test that verifies that the name show up in error messages. For example:

```python
try:
    Schema({"a": int}, name="MySchema").validate({"a": "1"})
except SchemaError as e:
    assert str(e) == "the expected message"
```

skorokithakis:
Yes, please run `pip install pre-commit; pre-commit install; pre-commit run -a`.

SnapperGee:
Added test for prepend_schema_name() function and used pre-commit

SnapperGee:
Just a heads up, but as long as there've been no changes to the upstream schema repo, this fork should be good to merge

skorokithakis:
I'm good to merge if @rmorshea is.

rmorshea:
LGTM 👍

skorokithakis:
Fantastic! Thank you, persons.

rmorshea:
Congrats @SnapperGee on your first open source contribution!

abbas-aghamirzaei:
hi.
how can I disable type checking(for performance)?
I want to only check the structure of dictionary. Is it possible?
thanks.

PR #184, opened 2018-11-29
Use correct singular/plural form of “key(s)” in error messages

jrosdahl:
Also added a new test case for covering the plural case of SchemaMissingKeyError.


skorokithakis:
Thank you!

PR #186, opened 2018-12-05
allow Optional(..., default=obj) to be a callable

jsurloppe:
allow optional to take a callable as default  (dict or list for example)


jsurloppe:
done, fixed pep8 too.

skorokithakis:
Thank you!

PR #202, opened 2019-04-01
Set long_description_content_type

Code0x58:
This is so things like [pypi.org](https://pypi.org/project/schema/) don't have to guess how to render the package description.


skorokithakis:
Thank you!

PR #206, opened 2019-05-21
Improve JSON schema generation

dblanchette:
This is adding on top of #180 

New features are:
- Regular expressions (e.g. `Regex("^something-.*")` will generate `{"type": "string", "pattern": "^something-.*"}`)
- const (e.g. `[1]` will generate `{"const": 1}`)
- enum (e.g. `["foo", "bar"]` will generate `{"enum": ["foo", "bar"]}`)
- And (e.g. `And(str, "test2")` will generate `{"allOf": [{"type": "string"}, {"const": "test2"}]}`)

I've also added the title of the source Schema in the generated JSON schema.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -537,7 +537,5 @@ Here's an example:
 Please note that this is a beta feature. Some JSON schema features are not implemented. Some caveats:
 
 - There are no object references, items of type `object` are always fully rendered
-- Some JSON schema types are not implemented. In those cases, an empty dict will be rendered.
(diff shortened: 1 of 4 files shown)


skorokithakis:
Looks good to me! Let me know if it's ready to merge.

dblanchette:
Thanks for the quick reply!

Ready to merge on my end

skorokithakis:
Merged, thank you!

PR #208, opened 2019-05-23
Feature/json schema descriptions

jcbedard:
Another feature for #180 
 - Added the Literal type for adding a description to JSON schemas
 - Fixed new bugs introduced by the Literal type
 - Added tests to ensure the Literal type was working as expected 

This is a direct continuation to PR #206

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,8 @@ matrix:
       dist: xenial
       sudo: true
     - env: TOXENV=coverage
-    - env: TOXENV=pep8
+    - env: TOXENV=checks
+      python: 3.6
 
 cache:
   pip: true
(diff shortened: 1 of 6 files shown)


skorokithakis:
Hmm, why so many changes here? Looks like most of it is reformatting?

jcbedard:
unfortunately black didn't like my formatting which was failing travis CI so I ran it through black

skorokithakis:
I'm confused, where did black not like your formatting? We should standardize on black, I think, so the code will look consistent from now on.

jcbedard:
I didn't know you guys used black with line length 120 so it created huge formatting changes, thus I committed it again with line 120 and now as you see everything passed

skorokithakis:
I have created #209 so this won't happen again, hopefully.

skorokithakis:
Is the new `Literal` type only intended to be used in JSONSchema? Or is there another use?

jcbedard:
only to be used in JSONSchema right now, this is most useful when using autocomplete, a description appears next the the key name

skorokithakis:
In that case, looks good to me!

PR #210, opened 2019-05-23
Add references to JSON schema rendering

dblanchette:
To reduce the size of the generated JSON schema, this PR adds setting an id for a part of the schema and then using the id as reference in other parts.

For our particular use case, the produced JSON schema decreased size from 54 MiB to 65 KiB using this feature.

Some caveats:
- The resulting schema is not the smallest it can be because of the way the tree is traversed. I tested different approaches to reduce it further, but thought that the added complexity and execution time was not worth the size reduction. In my above example, the size went down to 28 KiB with breadth-first traversal instead of 65 KiB with the implementation in this PR, still not a significant gain from the 54 MiB when not using references.
- This produces a single schema, not several files with references, [as is suggested here](https://json-schema.org/understanding-json-schema/structuring.html#reuse)
- The resulting schema is harder to read by a human. The id are hashes, not friendly names and the references are hard to follow. I don't think this is a big problem since this is more for machine consumption

I hesitated to set the `no_refs` parameter to True by default, since the resulting schema will be different (but still usable) from the previous versions. Is this a breaking change?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -509,7 +509,7 @@ Here's an example:
     >>> s = Schema({"test": str,
     ...             "nested": {Optional("other"): str}
     ...             })
-    >>> json_schema = json.dumps(s.json_schema("https://example.com/my-schema.json", no_refs=True))
+    >>> json_schema = json.dumps(s.json_schema("https://example.com/my-schema.json"))
 
     # json_schema
     {
@@ -534,8 +534,8 @@ Here's an example:
         "$schema":"http://json-schema.org/draft-07/schema#"
     }
 
-By default, in order to minimize the size of the output, the generated schema will use references to other parts of the schema. You can disable
-this behaviour by providing the parameter `no_refs` to the json_schema method.
-
 Please note that not all JSON schema validations are implemented. This includes features such as integers' minimum and maximum or
 arrays' minItems.
+
+In order to minimize the size of the output, the generated schema can be made to use references to other parts of the schema.
+Enable this behaviour by providing the parameter `use_refs` to the json_schema method.
--- a/README.rst
+++ b/README.rst
@@ -496,7 +496,7 @@ this is how you validate it using ``schema``:
 As you can see, **schema** validated data successfully, opened files and
 converted ``'3'`` to ``int``.
 
-(Beta feature) Generating JSON schema
+Generating JSON schema
 -------------------------------------------------------------------------------
 You can also generate standard `draft-07 JSON schema <https://json-schema.org/>`_ from a dict `Schema`.
 This can be used to add word completion and validation directly in code editors.
@@ -509,7 +509,7 @@ Here's an example:
     >>> s = Schema({"test": str,
     ...             "nested": {Optional("other"): str}
     ...             })
-    >>> json_schema = json.dumps(s.json_schema("https://example.com/my-schema.json"))
+    >>> json_schema = json.dumps(s.json_schema("https://example.com/my-schema.json", no_refs=True))
 
     # json_schema
     {
@@ -530,12 +530,12 @@ Here's an example:
             "nested"
         ],
         "additionalProperties":false,
-        "id":"https://example.com/my-schema.json",
+        "$id":"https://example.com/my-schema.json",
         "$schema":"http://json-schema.org/draft-07/schema#"
     }
 
-Please note that this is a beta feature. Some JSON schema features are not implemented. Some caveats:
+By default, in order to minimize the size of the output, the generated schema will use references to other parts of the schema. You can disable
+this behaviour by providing the parameter `no_refs` to the json_schema method.
 
-- There are no object references, items of type `object` are always fully rendered
-- Validations other than type are not implemented. This includes features such as integers'
-  minimum and maximum or arrays' minItems
+Please note that not all JSON schema validations are implemented. This includes features such as integers' minimum and maximum or
+arrays' minItems.
--- a/schema.py
+++ b/schema.py
@@ -291,6 +291,10 @@ class Schema(object):
     def description(self):
         return self._description
 
+    @property
+    def ignore_extra_keys(self):
+        return self._ignore_extra_keys
+
     @staticmethod
     def _dict_key_priority(s):
         """Return priority for a given key object."""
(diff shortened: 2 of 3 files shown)


skorokithakis:
> I hesitated to set the no_refs parameter to True by default, since the resulting schema will be different (but still usable) from the previous versions.

A clarification, as I'm not very familiar with JSONSchema, `no_refs=False` basically uses the old, more human-readable code? If so, I'm for setting it to `True` by default since I can see people wanting the shorter format, since the JSONSchema export is more usually intended for consumption by machines, as far as I know

> Is this a breaking change?

Good question, in reality every change is a breaking change, since users sometimes come to rely even on bugs. Since this is a beta feature, I'm okay with making this particular breaking change, if my understanding above is correct. I don't think many people will get bitten by this, since, in my understanding, the schemas are mostly equivalent, just less readable?

dblanchette:
> A clarification, as I'm not very familiar with JSONSchema, `no_refs=False` basically uses the old, more human-readable code?
`no_refs=False` means that we use refs, so it's the opposite.

I think that this is a confusing double negation, so I changed it to `use_refs`.

I thought about it during the weekend and I now think that using refs should not be on by default since it's only useful for very big schemas and it is not supported by all libraries and IDE tools. I changed that in my last commit.

I also added tests that will maybe explain a little bit more of the motivation behind the change.

Let's say I have this schema:
```
key_names = ["a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t"]
Schema({
        Literal(Or(*key_names), description="A key that can have many names"): {
            "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": str
        }
    })
```
The JSON schema would be:
```
{
    "required": [],
    "additionalProperties": false,
    "$id": "my-id",
    "$schema": "http://json-schema.org/draft-07/schema#",
    "type": "object",
    "properties": {
        "a": {
            "type": "object",
            "properties": {
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": {
                    "type": "string"
                }
            },
            "required": [
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references"
            ],
            "additionalProperties": false
        },
        "b": {
            "type": "object",
            "properties": {
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": {
                    "type": "string"
                }
            },
            "required": [
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references"
            ],
            "additionalProperties": false
        },
        {} // Repeated for keys c to t
    }
}
```

When using references:
```
{
    "$id": "my-id",
    "$schema": "http://json-schema.org/draft-07/schema#"
    "required": [],
    "additionalProperties": false,
    "type": "object",
    "properties": {
        "a": {
            "type": "object",
            "properties": {
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": {
                    "type": "string",
                    "$id": "#-5047904004637793631"
                }
            },
            "required": [
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references"
            ],
            "additionalProperties": false
        },
        "b": {
            "type": "object",
            "properties": {
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": {
                    "$ref": "#-5047904004637793631"
                }
            },
            "required": [
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references"
            ],
            "additionalProperties": false,
            "$id": "#-9158246229569689311"
        },
        "c": {
            "$ref": "#-9158246229569689311"
        },
        "d": {
            "$ref": "#-9158246229569689311"
        },
        "e": {
            "$ref": "#-9158246229569689311"
        },
        "f": {
            "$ref": "#-9158246229569689311"
        },
        "g": {
            "$ref": "#-9158246229569689311"
        },
        "h": {
            "$ref": "#-9158246229569689311"
        },
        "i": {
            "$ref": "#-9158246229569689311"
        },
        "j": {
            "$ref": "#-9158246229569689311"
        },
        "k": {
            "$ref": "#-9158246229569689311"
        },
        "l": {
            "$ref": "#-9158246229569689311"
        },
        "m": {
            "$ref": "#-9158246229569689311"
        },
        "n": {
            "$ref": "#-9158246229569689311"
        },
        "o": {
            "$ref": "#-9158246229569689311"
        },
        "p": {
            "$ref": "#-9158246229569689311"
        },
        "q": {
            "$ref": "#-9158246229569689311"
        },
        "r": {
            "$ref": "#-9158246229569689311"
        },
        "s": {
            "$ref": "#-9158246229569689311"
        },
        "t": {
            "$ref": "#-9158246229569689311"
        }
    }
}
```
Of course, a human being would write something more legible like
```
{
    "$id": "my-id",
    "$schema": "http://json-schema.org/draft-07/schema#"
    "required": [],
    "additionalProperties": false,
    "type": "object",
    "properties": {
        "a": {
            "$id": "#long_property",
            "type": "object",
            "properties": {
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references": {
                    "type": "string
                }
            },
            "required": [
                "a_very_long_name_that_i_dont_want_to_repeat_a_million_times_so_i_will_use_references"
            ],
            "additionalProperties": false
        },
        "b": {
            "$ref": "#long_property"
        },
        "c": {
            "$ref": "#long_property"
        },
        "d": {
            "$ref": "#long_property"
        },
        "e": {
            "$ref": "#long_property"
        },
        "f": {
            "$ref": "#long_property"
        },
        "g": {
            "$ref": "#long_property"
        },
        "h": {
            "$ref": "#long_property"
        },
        "i": {
            "$ref": "#long_property"
        },
        "j": {
            "$ref": "#long_property"
        },
        "k": {
            "$ref": "#long_property"
        },
        "l": {
            "$ref": "#long_property"
        },
        "m": {
            "$ref": "#long_property"
        },
        "n": {
            "$ref": "#long_property"
        },
        "o": {
            "$ref": "#long_property"
        },
        "p": {
            "$ref": "#long_property"
        },
        "q": {
            "$ref": "#long_property"
        },
        "r": {
            "$ref": "#long_property"
        },
        "s": {
            "$ref": "#long_property"
        },
        "t": {
            "$ref": "#long_property"
        }
    }
}
```

[You can read about $id and $ref here](https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref)

The diff may seem big at first, but it's not that many changes. I changed the `json_schema` method so that it has a recursive inner method since I needed to outer method to store the `seen` dict that contains the hash of the dict already in the schema. The other change is making the method check for an existing reference before returning.

skorokithakis:
Looks good to me, thanks!

PR #211, opened 2019-05-31
Fix error in README.rst

nedbat:
I think this might make the readme render properly on PyPI

Before this fix, Sphinx says: "WARNING: Error in "code" directive: maximum 1 argument(s) allowed, 18 supplied."


skorokithakis:
Oops, thank you!

PR #212, opened 2019-06-03
JSON Schema: Fix using Literal in enum

dblanchette:
Fixes a bug introduced by #208.

The Literal type could not be rendered in JSON because it could not be cast to string. Moreover, the implementation of Or assumed the keys were string, but they could be Literal.

Sorry for the bug!


skorokithakis:
No problem, thanks for the fix!

PR #213, opened 2019-06-18
JSON schema with common definitions + Update README

dblanchette:
Yet another JSON schema PR!

This one let users give a name to a part of their schema and use it as a reference. I was not satisfied with the previous implementation that would compute hashes and create a schema that is hard to read by humans. It was also not compatible with all IDEs. This is explained here: https://json-schema.org/understanding-json-schema/structuring.html#reuse

I also added a lot more information and exemples in the README, now every JSON schema feature is explained. Tell me if you think this could be improved.

By using this along with [docson](https://github.com/lbovet/docson) we now have a configuration schema that is light, works with most IDEs and is automatically documented.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -497,11 +497,17 @@ this is how you validate it using ``schema``:
 As you can see, **schema** validated data successfully, opened files and
 converted ``'3'`` to ``int``.
 
-Generating JSON schema
+JSON schema
 -------------------------------------------------------------------------------
+
 You can also generate standard `draft-07 JSON schema <https://json-schema.org/>`_ from a dict `Schema`.
-This can be used to add word completion and validation directly in code editors.
-Here's an example: 
+This can be used to add word completion, validation, and documentation directly in code editors.
+The output schema can also be used with JSON schema compatible libraries.
+
+JSON: Generating
+~~~~~~~~~~~~~~~~
+
+Just define your schema normally and call `json_schema` on it. The output is a Python dict, you need to dump it to JSON.
 
 .. code:: python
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -497,11 +497,17 @@ this is how you validate it using ``schema``:
 As you can see, **schema** validated data successfully, opened files and
 converted ``'3'`` to ``int``.
 
-Generating JSON schema
+JSON schema
 -------------------------------------------------------------------------------
```
skorokithakis: Can you make this the same width as the line above?

README.rst
```diff
@@ -497,11 +497,17 @@ this is how you validate it using ``schema``:
...
 You can also generate standard `draft-07 JSON schema <https://json-schema.org/>`_ from a dict `Schema`.
-This can be used to add word completion and validation directly in code editors.
-Here's an example: 
+This can be used to add word completion, validation, and documentation directly in code editors.
+The output schema can also be used with JSON schema compatible libraries.
+
+JSON: Generating
+~~~~~~~~~~~~~~~~
+
+Just define your schema normally and call `json_schema` on it. The output is a Python dict, you need to dump it to JSON.
```
skorokithakis: Probably better to say `call ``.json_schema()`` on it`.

skorokithakis:
I left some comments, I think it looks great otherwise, thank you for your help!

dblanchette:
Thanks for the review!

skorokithakis:
Looks great, thanks for the help!

PR #214, opened 2019-06-27
JSON Schema: Fix using `dict` as type

dblanchette:
I forgot to handle the "dict" type which would result in JSON schemas with "string" instead of "object" for parameters that are of the dict type, but that doesn't have keys specified.


skorokithakis:
Thank you!

PR #215, opened 2019-06-30
README: add 3.7 to list of tested versions

jim-obrien-orig:
Hi @keleshev ,
I'm very grateful for this lib. It's one of my favorite python packages - lightweight and speedy.
It seems like everything except for pre-commit `black` formatting is covered for `py37`, and it's definitely tested with 3.7.  Although I verified py37 so I can use on a current 3.7 project, it would be great if it were in the README. If there's a specific reason 3.7 isn't listed that's understandable.
Thanks again for the awesome library!


skorokithakis:
Thank you, we just forgot to add this!

PR #216, opened 2019-07-30
JSON Schema: Don't add a description in a ref

dblanchette:
Reading [the Reuse section of Understanding JSON Schema](https://json-schema.org/understanding-json-schema/structuring.html#reuse), I notice there should be nothing other than `$ref` when reusing another section of the schema.

This fixes the library behavior by adding the description on the referenced element instead.


skorokithakis:
Looks good, thanks!

PR #217, opened 2019-07-30
JSON Schema: Include default values

dblanchette:
When using Optional as a key, a default value can be set. With this change, the default value will be added to the generated JSON schema as well.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -620,6 +620,8 @@ class Schema(object):
                             expanded_schema[key_name] = _json_schema(
                                 sub_schema, is_main_schema=False, description=_get_key_description(key)
                             )
+                            if isinstance(key, Optional) and hasattr(key, "default") and key.default:
+                                expanded_schema[key_name]["default"] = key.default
                         elif isinstance(key_name, Or):
                             # JSON schema does not support having a key named one name or another, so we just add both options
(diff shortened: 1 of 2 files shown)


dblanchette:
@skorokithakis Wow, that was merged really fast. Thanks!

skorokithakis:
Thank *you*!

dblanchette:
Hi @skorokithakis !

I was wondering if you were planning to make a new release that includes the JSON schema changes. We have been using our fork for a while, but it would be nice to go back to the official repository.

skorokithakis:
Hey @dblanchette! Certainly, has the fork worked well for you? I haven't tested the JSONschema stuff at all, but if it's relatively stable I can cut a release.

dblanchette:
Actually, I've just found a bug with this PR: default values are not added if they evaluate to False. I will submit a PR for the fix later today.

We have been using JSON schemas generated from this library for a while here at Coveo to document our deployment pipeline configuration file and it has served us well.

I think a release after my next fix would be appropriate, so more people can benefit from it :)

As a side note, I recently created a jinja2 template to generate a static documentation website from the resulting schema, which I will open source soon. That is the reason I added the default values to the schema.

skorokithakis:
Oh that sounds fantastic, thanks! I'll be waiting for the fix then.

PR #218, opened 2019-08-01
JSON Schema: Fix using falsy values as default

dblanchette:
There's actually 2 different fixes here

First, default values were not added to the JSON Schema if they were falsy (`None`, `False`, `[]`, `{}`, `0` are correct default values). Since `key.default` is not set on `Optional` if it has no default, there was no need to check the value anyway.

When fixing that, I found a bigger problem. Values used in both `const` and `default` were simply cast to string. You would end up with
```json
{
    "const": "None"
}
```

instead of

 ```json
{
    "const": null
}
```
It will now make an attempt to convert the value to a JSON serializable type but fallback to casting as a string.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -507,6 +507,19 @@ class Schema(object):
                     return "object"
                 return "string"
 
+            def _to_json_type(value):
+                """Attempt to convert a constant value (for "const" and "default") to a JSON serializable value"""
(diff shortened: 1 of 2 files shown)


skorokithakis:
Looks good to me, thanks! I will also release soon, but feel free to bug me if I forget!

Issue #219, opened 2019-09-04
Types inside `And` are ignored when creating JSON schema


berkanteber:
Currently, it is as follows:

```
>>> Schema({'name': str}).json_schema("example_schema")
{
    'type': 'object',
    'properties': {'name': {'type': 'string'}},
    'required': ['name']
    'additionalProperties': False,
    'id': 'example_schema',
    '$schema': 'http://json-schema.org/draft-07/schema#'
}

>>> Schema({'name': And(str, Use(str.lower))}).json_schema("example_schema")
{
    'type': 'object',
    'properties': {'name': {}},
    'required': ['name']
    'additionalProperties': False,
    'id': 'example_schema',
    '$schema': 'http://json-schema.org/draft-07/schema#'
}
```

Converting `str` to `And(str, Use(str.lower))` results in dropping the `{'type': 'string'}` in JSON schema. However, if there is a single type in `And`, it should be preserved in JSON schema.

skorokithakis:
Thanks for reporting! @dblanchette?

dblanchette:
I'll take a look

dblanchette:
@berkanteber Using the version from master, I get

```>>> json.dumps(Schema({'name': And(str, Use(str.lower))}).json_schema("example_schema"), indent=4)

{
    "type": "object",
    "properties": {
        "name": {
            "allOf": [
                {
                    "type": "string"
                }
            ]
        }
    },
    "required": [
        "name"
    ],
    "additionalProperties": false,
    "$id": "example_schema",
    "$schema": "http://json-schema.org/draft-07/schema#"
}```

While this is not ideal (the "allOf" is not necessary if there is only one possibility), at least the schema is valid. I'll check to fix that.

dblanchette:
@skorokithakis Do you think it would be time to make a release so that people can benefit from the new feature and fixes for JSON Schema?

skorokithakis:
Yes, if master is stable enough, we definitely could. Have you used `master` much, @dblanchette?

dblanchette:
We've been using the master of our fork (identical to master here) for creating JSON schemas for several weeks without major issues. The minor issues were fixed along the way.

The schema we are generating is pretty complex with a lot of conditions so I think it is a good candidate.

skorokithakis:
Fantastic, I will make a release very soon (but feel free to bug me if I forget).

skorokithakis:
0.7.1 released!

dblanchette:
[...]

PR #220, opened 2019-09-04
JSON Schema: Fix allOf and oneOf with only one condition

dblanchette:
Fixes an issue raised while investigation #219


skorokithakis:
Thank you! I can merge when tests pass.

PR #222, opened 2019-11-02
Relax contextlib2 requirement

tobim:
0.5.5 is not the latest version any more.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-contextlib2==0.5.5
+contextlib2>=0.5.5


skorokithakis:
Is there any problem that is solved with this upgrade?

tobim:
This is purely for depending projects that also need a newer version of contextlib2 directly or indirectly through another library.

skorokithakis:
Ah, that's fair. Have you tested that schema works with the latest version of contextlib2?

tobim:
I just did and it does work. I did not add an upper bound because contextlib2 is a compatibility layer that _should_ not change existing functionality unless python proper does the same, in which case your code has to adapt anyway.

skorokithakis:
Alright, sounds reasonable, thanks!

pat-bert:
Would it make sense to release this as the latest version of schema so that one can benefit from it?

graingert:
@skorokithakis can you release this change on pypi please?

graingert:
@skorokithakis it would make sense as `0.7.1.patch0`

skorokithakis:
Certainly, I'll do that soon.

skorokithakis:
Released 0.7.2.

PR #228, opened 2020-04-15
.pre-commit-config.yaml: Add codespell, flake8, isort

cclauss:
Some functions are quite complex.  https://github.com/pycqa/mccabe  Could they be refactored?
```
black....................................................................Passed
codespell................................................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1
schema.py:341:5: C901 'Schema.validate' is too complex (30)
schema.py:459:5: C901 'Schema.json_schema' is too complex (57)
test_schema.py:239:1: C901 'test_dict' is too complex (17)
isort....................................................................Passed
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
       - id: flake8
         args:
           - --max-complexity=10
-          - --max-line-length=361  # 88
+          - --max-line-length=88
   - repo: https://github.com/timothycrosley/isort
     rev: 4.3.21-2
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,37 @@
 repos:
--   repo: https://github.com/ambv/black
-    rev: 18.9b0
+  - repo: https://github.com/python/black
+    rev: stable
     hooks:
-    - id: black
+      - id: black
+  - repo: https://github.com/codespell-project/codespell
+    rev: v1.16.0
+    hooks:
+      - id: codespell
+        args:
+          - --quiet-level=2
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.9
+    hooks:
+      - id: flake8
+        args:
+          - --max-complexity=10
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,37 @@
...
+    hooks:
+      - id: codespell
+        args:
+          - --quiet-level=2
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.9
+    hooks:
+      - id: flake8
+        args:
+          - --max-complexity=10
```
cclauss, line 17: This could be adjusted upwards but 10 is a good recommendation.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,37 @@
...
+          - --max-line-length=88
+  - repo: https://github.com/timothycrosley/isort
```
skorokithakis, line 19: I prefer https://github.com/asottile/reorder_python_imports rather than isort as the former reorders everything statically.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,37 @@
...
+    rev: 4.3.21-2
+    hooks:
+      - id: isort
+        args:
+          - --recursive
+  #- repo: https://github.com/pre-commit/mirrors-mypy
```
skorokithakis, line 25: I would rather not add commented-out sections, we can add this when we need it.

.pre-commit-config.yaml
```diff
@@ -1,5 +1,37 @@
...
     hooks:
-    - id: black
+      - id: black
+  - repo: https://github.com/codespell-project/codespell
+    rev: v1.16.0
+    hooks:
+      - id: codespell
+        args:
+          - --quiet-level=2
+  - repo: https://gitlab.com/pycqa/flake8
```
skorokithakis, line 12: This fails right now, and I don't want to make CI fail until we manage to refactor the functions. We should only include this along with other refactoring work so CI doesn't fail.

PR #230, opened 2020-04-27
JSON Schema: Add description next to $ref

dblanchette:
Starting with draft-04, the JSON schema specification allows having other attributes next to `$ref`. This is useful for us when using descriptions, because we may want to use a different description depending on where the sub schema is used.

Descriptions from `Literal` used as keys were already added to the value it references. With this change, descriptions will be added alongside the $ref node if the value is another schema and that schema has `as_reference` set to True.

For an example, look at `test_json_schema_definitions_and_literals`.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -533,10 +533,9 @@ class Schema(object):
             return_schema = {}
 
             is_a_ref = allow_reference and schema.as_reference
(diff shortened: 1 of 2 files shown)


skorokithakis:
This looks good, thanks!

PR #234, opened 2020-06-17
Do not drop previous errors within an Or criterion.

blaa:
When raising a SchemaError with a user readable error message this message would
be dropped if there was more than one validator in an Or() clause.

BTW: Thanks for the library, we use it extensively to keep our model correct.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -186,10 +186,9 @@ Now you can write your own validation-aware classes and data types.
 Lists, similar containers
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If ``Schema(...)`` encounters an instance of ``list``, ``tuple``, ``set`` or
-``frozenset``, it will validate contents of corresponding data container
-against schemas listed inside that container:
-
+If ``Schema(...)`` encounters an instance of ``list``, ``tuple``, ``set``
+or ``frozenset``, it will validate contents of corresponding data container
+against all schemas listed inside that container and aggregate all errors:
 
 .. code:: python
 
@@ -200,6 +199,7 @@ against schemas listed inside that container:
     Traceback (most recent call last):
     ...
     schema.SchemaError: Or(<class 'int'>, <class 'float'>) did not validate 'not int or float here'
+    'not int or float here' should be instance of 'int'
     'not int or float here' should be instance of 'float'
 
 Dictionaries
--- a/README.rst
+++ b/README.rst
@@ -20,7 +20,7 @@ entries with personal information:
 
 .. code:: python
 
-    >>> from schema import Schema, And, Use, Optional
+    >>> from schema import Schema, And, Use, Optional, SchemaError
 
     >>> schema = Schema([{'name': And(str, len),
     ...                   'age':  And(Use(int), lambda n: 18 <= n <= 99),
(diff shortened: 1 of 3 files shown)


skorokithakis:
Thank you for the PR!

skorokithakis:
Hmm, the tests are failing, could you look into that?

blaa:
Yeah, just checking. Locally they seems to be failing without my commit too:
```
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,8 @@
    -...
    -SchemaUnexpectedTypeError: '123' should be instance of 'int'
    +  File "/usr/lib/python3.8/doctest.py", line 1336, in __run
    +    exec(compile(example.source, filename, "single",
    +  File "<doctest README.rst[7]>", line 1, in <module>
    +    Schema(int).validate('123')
    +  File "/home/bla/_safe_sync/_projects/schema/schema.py", line 428, in validate
    +    raise SchemaUnexpectedTypeError(message, e.format(data) if e else None)
    +schema.SchemaUnexpectedTypeError: '123' should be instance of 'int'

/home/bla/_safe_sync/_projects/schema/README.rst:79: DocTestFailure
```
But I can see if I can fix it anyway. On travis it seems that python3 works ok, while python2 fails.

skorokithakis:
Oh hmm, I thought they were caused by the commit because they looked similar. If you could fix them, I would be grateful!

blaa:
Never worked with travis before, but let's see. ;)
Most seem to be caused by exceptions having the "schema." prefix now. I can change the docs to keep this, but I guess it would be cleaner to somehow change imports so that it's simpler.

skorokithakis:
Looking at that error, it *should* fail, as `'123'` isn't an int. I don't know whether the docs mean it to be failing, or if they mean to test for `123`...

blaa:
It passes (and stops and the next error) when I add "schema." prefix to the SchemaError exception. Except for this detail - the error matches.

blaa:
Hm. Ok, but that's because I test docs on python3. I haven't got python27

skorokithakis:
Hmm, for me they pass on master but fail on this branch.

blaa:
I have to learn to execute those doctests on python 2.7 and then do retest. Generic pytest on my py3 works.  It might take a while, but I'll try to fix those.

blaa:
I've migrated the docs to the python3. Hope it's ok? Maybe it's time? :) 
https://github.com/blaa/schema/tree/migrate-docs 
Not sure how to check travis without spamming pull requests.

Now I've got the error caused by my commit.

blaa:
Well. That should do it. But check if you accept the changes to the docs.

skorokithakis:
Oh yes, thank you, it's definitely time to move the docs to 3. I will merge this and check that docs generate fine, thanks again!

blaa:
I've noticed I've left the "SchemaError" imported in docs at the top - it's not strictly necessary. Sorry for this. ;) 
Learned a lot about travis and doctests though.

skorokithakis:
I can take that out, no problem, though it looks like `schema` refers to the instance of `Schema()` and not the module, which works by happy coincidence because the instance contains the error classes. It's also possible I've just misunderstood how doctests work :P

Thanks again!

blaa:
This "schema.SchemaError", as I understand the doctests, is only matching the output - not really relating to the local variables - and python3 names the exception more thoroughly.

skorokithakis:
Ah, great, thanks!

PR #237, opened 2020-06-26
JSON Schema: Support schemas where the root is not a dict

dblanchette:
The first version of the JSON Schema generation feature did not accept having any types other than a dict for the root.

This means that calling `json_schema` on `Schema({"test": [1, 2, 3]})` would work but on `Schema([1, 2, 3])`, it would just return the error "The main schema must be a dict."

After testing, turns out this was an artificial limitation and nearly no code change is needed to have it work with any supported types.

This will also make the examples in the README actually work :)


skorokithakis:
That's great, thank you!

PR #243, opened 2020-08-18
JSON Schema: Set additionalProperties true when dict contains str as key

dblanchette:
When a dict has a `str` or `object` as one of its keys, `schema` is accepting extra keys (or additional properties in JSON schema language). However, the generated JSON schema will have `additionalPropeties` set to False.

Currently, the only way to have `additionalPropeties` set to true is to use `ignore_extra_keys` in the `Schema` object constructor. Users may not want to do that because that means that `validate` will not include the keys in the result.

This fix will have `additionalPropeties` set to true if `ignore_extra_keys` is True or if any of the key is `str` or `object` (or `Optional(str)` or `Optional(object)`).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -599,7 +599,7 @@ Implemented
 
     ``{'type': 'object', 'properties': {'test': {'type': 'string'}}, 'required': [], 'additionalProperties': False}``
 
-    additionalProperties is set to true when at least one of the condition is met:
+    additionalProperties is set to true when at least one of the conditions is met:
         - ignore_extra_keys is True
         - at least one key is `str` or `object`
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -599,6 +599,26 @@ Implemented
 
     ``{'type': 'object', 'properties': {'test': {'type': 'string'}}, 'required': [], 'additionalProperties': False}``
 
+    additionalProperties is set to true when at least one of the condition is met:
```
skorokithakis, line 602: "conditions" here.

skorokithakis (changes requested):

skorokithakis:
Looks good to me, I just have a minor typo change request.

skorokithakis:
Thank you!

PR #244, opened 2020-09-01
fix: JSON Schema: Allow using $ref when schema is not a dict

dblanchette:
Similarly to #243, using `as_reference` only worked when the schema is an object (dict).

This change seems longer than it is, it is mainly just moving code. The part creating a $ref object has been moved to before checking the flavor, instead of being inside the `flavor == DICT` section.

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -533,80 +533,77 @@ class Schema(object):
 
             return_schema = {}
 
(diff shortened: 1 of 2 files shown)


skorokithakis (approved):

skorokithakis:
Looks good to me, thank you!

PR #245, opened 2020-10-01
Fix "Unknown format code" in Python 3.8

dblanchette:
On Python 3.8.6, we have started getting this weird error:
```ValueError: Unknown format code 'b' for object of type 'str'```

Turns out there was a change in Python itself so that it is not possible to use a format string with a `RegexFlag` object anymore.

Casting to int resolves the issue while keeping compatibility with older Python versions.


skorokithakis:
Thank you!

Issue #241, opened 2020-08-03
General changes to error handling in schema


kmaork:
## Problem
In my opinion, the way `SchemaError`s are produced can be improved. There are a few problems with the current implementation:
1. Error string formatting logic is duplicated throughout the code, does not exist in all needed cases (e.g. callable validators).
2. `SchemaError` objects do not contain a reference to the data that validation failed on, which could be useful to users catching the exception.
3. Many functions throw `SchemaError`s in a context in which a `SchemaError` is handled, which causes a long traceback containing `while handling this exception, another exception has occured: ...`. This stack trace doesn't contain any useful information because we collect the `autos` and `errors` from the handled `SchemaError` exceptions anyway.
4. `SchemaError`'s interface is hard to understand at first (mainly because of `autos` and `errors`) for anyone wanting to implement a custom validator that can throw a `SchemaError`'.

## Possible solution
I would like to offer a new implementation of  `SchemaError` that will hopefully solve all of the above. Pseudocode:
```python
# The new SchemaError class
class SchemaError(Exception):
    def __init__(self, generated_error_msg, user_error_msg, invalid_data):
        super(Schema, self).__init__()
        self.generated_error_msg = generated_error_msg
        self.user_error_msg = user_error_msg
        self.invalid_data = invalid_data
        
    def __str__(self):
        return '{}: {}'.format(self.generated_error_msg, self.user_error_msg.format(self.invalid_data))

# Usage example
try:
    s.validate(data)
except SchemaError as e:
    six.raise_from(SchemaError(auto, user_msg, data), e)
```
This solves the problems above:
1. Formatting is handled in one place.
2. The exception contains the data.
3. We don't collect  `autos` and `errors` anymore, and let python's exception chaining do the job instead.
4. IMO the new interface is much simpler and more pythonic.

## Drawbacks
Some users might have `raise SchemaError(...)` in their code. Such code will break if we change the interface of `SchemaError.__init__`. We could:
- Change the interface and break compatability (maybe bump a major version)
- Try to make the new interface backwards compatible - maybe keep the `autos` and `errors` arguments and have the `invalid_data` argument default to None.
-  We can make a new class for the proposed exception, and have it inherit from `SchemaError`. That way, `SchemaError` could still be raised, and catching `SchemaError`s will catch the new exception type as well.

skorokithakis:
[...]

PR #247, opened 2020-11-25
don't double-format errors. fixes #240

### LINKED ISSUE #240: Formatting error when supplying raw dict as schema to `Or`
I'm not sure if this is valid Schema usage, but I wanted to drop my reproduction here just in case.

I have a schema that looks like this:

```python
  import schema as s

  test1 = s.Or(str, {
      s.Optional("gpu"): str,
      s.Optional("cpu"): str
  },
               error="Got '{}'")

  test1.validate({"cpu": "face", "gpu": "cake"})
```

Trying to run that code throws this error:

```python
self = Schema(<class 'dict'>), data = {'cpu': 'face', 'gpu': 'cake'}

    def validate(self, data):
      Schema = self.__class__
      s = self._schema
>     e = self._error.format(data) if self._error else None
E     KeyError: "'cpu'"

env/lib/python3.6/site-packages/schema.py:372: KeyError
```

When I trace, it looks like what's happening is that `self._error` on the `Or` schema is getting set to the FILLED schema, on this line:

https://github.com/keleshev/schema/blob/master/schema.py#L345

ie, when I insert `print(self._error)` there, I see:

```python
Got '{}'
Got '{'cpu': 'face', 'gpu': 'cake'}'
```

Which creates this case:

```python
>>> "Got '{'cpu': 'face', 'gpu': 'cake'}'".format({'cpu': 'face', 'gpu': 'cake'})
KeyError: "'cpu'"
```

The issue's solved if I wrap the dictionary in `s.Schema`... but it might be worth having `Or`, `And` etc do the wrapping for us if it sees a raw dict to prevent this issue.

Hope this helps!

leif:
Pull request #238 introduced a change (which was released in 0.7.3) which
caused errors to sometimes be formatted twice, which caused valid data to
sometimes appear to be invalid, as described in issue #240.

This commit reverts pull request #238, which fixes issue #240, and also adds
error formatting to some other exceptions including the one which gets raised
in the example described as motivation for the change in #238.

It also adds test cases to ensure the behaviors desired in both #238 and #240
remain present.

This commit is also relevant to anyone looking at issue #241 (though it does
not implement the plan described there).

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -220,7 +220,7 @@ class Regex(object):
             if self._pattern.search(data):
                 return data
             else:
-                raise SchemaError("%r does not match %r" % (self, data), e)
+                raise SchemaError("%r does not match %r" % (self, data), e.format(data) if e else None)
         except TypeError:
             raise SchemaError("%r is not string nor buffer" % data, e)
 
@@ -344,7 +344,7 @@ class Schema(object):
     def validate(self, data):
         Schema = self.__class__
         s = self._schema
-        e = self._error.format(data) if self._error else None
+        e = self._error
         i = self._ignore_extra_keys
 
         if isinstance(s, Literal):
@@ -397,7 +397,7 @@ class Schema(object):
                                 except SchemaError as x:
                                     k = "Key '%s' error:" % nkey
                                     message = self._prepend_schema_name(k)
-                                    raise SchemaError([message] + x.autos, [e] + x.errors)
+                                    raise SchemaError([message] + x.autos, [e.format(data) if e else None] + x.errors)
                                 else:
                                     new[nkey] = nvalue
                                     coverage.add(skey)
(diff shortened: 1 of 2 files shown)


ioerror:
@skorokithakis Would it be possible to merge this PR?

skorokithakis (approved):

skorokithakis:
Certainly, sorry, not sure why I didn't see this. Does it work well for you, @ioerror? Have you tried it?

ioerror:
> Certainly, sorry, not sure why I didn't see this. Does it work well for you, @ioerror? Have you tried it?

Yeah, @leif and I would like this to land in a tagged release and hopefully make it into Debian before the next package freeze. We're using this (module) for a novel cryptographic protocol and it is really extremely helpful.

Would it be possible to tag a new release of schema that includes this change set? I think that might make it possible for the Debian packagers to package it, if they want to ship an updated version for Bullseye, that is.

skorokithakis:
@ioerror sure, why not. Will do shortly, thanks!

skorokithakis:
@ioerror Released 0.7.4, please let me know if anything with it is wrong.

ioerror:
Thank you! We will test things and one of us will report back.

Issue #265, opened 2021-07-22
Derived class from `Optional` is not allowed


raoyitao:
I have a validator inherit class `Optional`, however I found that it is not allowed: https://github.com/keleshev/schema/blob/master/schema.py#L420

From this line **set(k for k in s if type(k) is Optional and hasattr(k, "default"))** I saw that only `Optional` object can be handled, is it possible to use **isintance(k, Optional)** instead, so that derived class of `Optional` can work. `Hook` can be inherited, it will be nice that `Optional` can also be inherited.

raoyitao:
another piece of code at https://github.com/keleshev/schema/blob/master/schema.py#L320

`@staticmethod
def _is_optional_type(s):
    """Return True if the given key is optional (does not have to be found"""
    return any(isinstance(s, optional_type) for optional_type in [Optional, Hook])`

`_is_optional_type` only checks if the schema is a instance of `Optional`, so a subclass instance can also work.

PR #268, opened 2021-09-23
Full subclass support for nested schemas.

### LINKED ISSUE #262: A post validation hook at the schema level would be handy.
Sometimes there's some validation, default or restructure that I'd like to do after parsing. I can do this by walking the structure at the end, but it'd be nice if I could pass a callable to a Schema arg that let me mess with the result. Somewhat similar to post_init in dataclasses. The big advantage of this over doing it myself is I can attach it at a relevant point within a nested structure.

A small example that isn't dumb is hard, but instead of this

```
events_schema = schema.Schema(
    {
        str: {
            "capacity": int,
            "minimum": int,  # default to capacity
        }
    }
)

events = events_schema.validate(data)

for event in events.items():
    if "minimum" not in event:
        event["minimum"] = event["capacity"]
```

you would be able to have

```
def default_minimum(event):
    if "minimum" not in event:
        event["minimum"] = event["capacity"]

events_schema = schema.Schema(
    {
        str: Schema(
            {
                "capacity": int,
                "minimum": int,  # default to normal_capacity
            }, 
            post_validate = default_minimum,
        )
    }
)

events = events_schema.validate(data)
```

Particularly note how the loop went away, that's gets nicer the more nested the schema is.

Jerry-Ma:
Hi,

I have encounter some issues in integrating this package with the python's new dataclasses.dataclass. Most of them is due to the lack of customizability of the core classes including `Schema`, `Optional`, etc.

This pull request contains the following changes:

* Added `**kwargs` to the function signature of `Schema.validate()`. This allows subclass to override the behavior of `validate` in a propagate-able fashion. Nested `schema.Schema` class will ignore any kwargs, but nested `MySchema` class will behave accordingly. This fix #262 (@tolomea). Moreover, the kwargs passed to `validate()` is picked up by the `Optional(default=)` value when it is a callable:
```
assert Schema({Optional('a', default=lambda **k: k['some_value'])}).validate({}, some_value=1) == {'a': 1}
```

* Fix the hardcoded type check for `Optional` mentioned in #265 (@raoyitao). This allows one to implement MyOptional that consumes the kwargs passed from `validate` to callable default more integrated:

```
class MyOptional(Optional):
    @property
    def default(self):
        def wrapped(**k):
            return k['some_value']
        return wrapped
    @default.setter
    def default(self, _):
        pass # this is needed to satisfy the Optional.__init__ which does `self.default=default`
``` 

* Added `**kwargs` to the signature of `Schema.json_schema()`, and changed the handling of default values such that callable will be evaluated with the kwargs. This makes it more inline with the behavior of the `Schema.validate`:

```
    def default_func(**kwargs):
        return 'Hello, ' + kwargs['name']
    s = Schema({Optional("test", default=default_func): str})
    assert s.json_schema("my-id", name='World!') == {
        "$schema": "http://json-schema.org/draft-07/schema#",
        "$id": "my-id",
        "properties": {"test": {"default": "Hello, World!", "type": "string"}},
        "required": [],
        "additionalProperties": False,
        "type": "object",
    }
```

For more details of the changes and their implications, please take a look at the test cases I added.
  
Please let me know what you think!

### DIFF
--- a/schema.py
+++ b/schema.py
@@ -459,7 +459,7 @@ class Schema(object):
             message = self._prepend_schema_name(message)
             raise SchemaError(message, e.format(data) if e else None)
 
-    def json_schema(self, schema_id, use_refs=False):
+    def json_schema(self, schema_id, use_refs=False, **kwargs):
         """Generate a draft-07 JSON schema dict representing the Schema.
         This method must be called with a schema_id.
 
@@ -650,7 +650,7 @@ class Schema(object):
                                 sub_schema, is_main_schema=False, description=_get_key_description(key)
                             )
                             if isinstance(key, Optional) and hasattr(key, "default"):
-                                expanded_schema[key_name]["default"] = _to_json_type(key.default() if callable(key.default) else key.default)
+                                expanded_schema[key_name]["default"] = _to_json_type(key.default(**kwargs) if callable(key.default) else key.default)
                         elif isinstance(key_name, Or):
                             # JSON schema does not support having a key named one name or another, so we just add both options
                             # This is less strict because we cannot enforce that one or the other is required
--- a/schema.py
+++ b/schema.py
@@ -117,7 +117,7 @@ class And(object):
         """The provided parameters"""
         return self._args
 
-    def validate(self, data):
+    def validate(self, data, **kwargs):
         """
         Validate data using defined sub schema/expressions ensuring all
         values are valid.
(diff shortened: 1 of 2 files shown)


skorokithakis:
This seems great, thank you! I especially appreciate the detailed PR description. Could you add some documentation to the README about these features so we can merge?

Jerry-Ma:
@skorokithakis

Sorry for the long wait. I've just updated the README with a short example demonstrate this feature.

PR #273, opened 2021-12-30
Drop support for obsolete python versions

sevdog:
Drops support for Python < 3.6 (also removing requirements for `contextlib2` backport).

Also adds support for Python 3.10


skorokithakis:
Hmm, I'm not opposed to this, but why drop old versions? What does it achieve?

sevdog:
> Hmm, I'm not opposed to this, but why drop old versions? What does it achieve?

Since old versions of python are no more developer nor supported if there is no a founded request to support them it is cheaper to drop them. Expecially for python 2.x which [has sunset on 1 Jan 2020](https://www.python.org/doc/sunset-python-2/), and brings a lot o compatibility issues (like "futures", string types detection and backports).

This can also be achieved at code level using six and backports, but the paingoes beyond the gain IMO (expecially if you want to keep also Python 2.6 which reached [EOL in 2013](https://www.python.org/downloads/release/python-269/)).

This may be related to a "major" release, since may introduce "backwards incompatible changes" (even if is just an issue related to the interpreter).

skorokithakis:
I agree in general, but would rather couple this with some breaking changes (ie drop support for old versions when we use new features). On one hand, if we don't drop support, nobody will know they can use the new features. On the other hand, if we drop support now and need to release a small fix or feature soon, that'll have to be a major version...

gschaffner:
Closes #282.

gschaffner:
Resolves the test regression on Python < 3.3 caused by #268; see #276.

gschaffner:
@skorokithakis Maybe I misunderstand, but I don't think that a small fix/feature soon would necessitate a major version bump if this was merged.

Under SemVer, dropping support for old versions of Python can be done as a minor version bump. Old Pythons can be dropped in a backwards-compatible way if, in packaging, you prevent the new `schema` version from being installed on Pythons it does not support. For setuptools via `setup.py`, passing `python_requires=">=3.6"` would do this.

skorokithakis:
Hmm, come to think of it, you're right, @gschaffner. We'd need to make these changes in the `setup.py`, but it sounds like that'd be OK.

PR #276, opened 2022-03-05
Fix `Optional` failing when `default` is some builtins, and fix incompatibility with Python 2.7

### LINKED ISSUE #275: `Optional` fails when `default` is some builtins (e.g. dict, set, list)
The documentation provides the example

https://github.com/keleshev/schema/blob/09c00eda9599e53f7e6b84d7c91ecd3b42f71772/README.rst#L271-L277

This does not work, however:

```python
>>> from schema import Schema, Optional 
>>> Schema({Optional('data', default=dict): {}}).validate({})
  File "/tmp/tmp.27EtQAmlOm/.venv/lib/python3.10/site-packages/schema.py", line 431, in validate
    new[default.key] = _invoke_with_optional_kwargs(default.default, **kwargs) if callable(default.default) else default.default
  File "/tmp/tmp.27EtQAmlOm/.venv/lib/python3.10/site-packages/schema.py", line 277, in _invoke_with_optional_kwargs
    s = inspect.signature(f)
  File "/usr/lib/python3.10/inspect.py", line 3247, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped,
  File "/usr/lib/python3.10/inspect.py", line 2995, in from_callable
    return _signature_from_callable(obj, sigcls=cls,
  File "/usr/lib/python3.10/inspect.py", line 2529, in _signature_from_callable
    raise ValueError(
ValueError: no signature found for builtin type <class 'dict'>
```

I've tested and found the above error to occur with schema v0.7.5 on CPython 3.10.2, CPython 3.7.12, and PyPy3.8-7.3.7.

It looks like this regression was introduced in 958e1ad657273537e73c431fae356ae15eff45c9. Arguably this is due to a bug in Python, though.

Curiously: `inspect.signature(list)` works fine in CPython 3.6.15, but raises in 3.7.0.

For users wanting to work around this, some working options include e.g. `Optional(..., default=lambda: dict())` and `Optional(..., default=dict.__call__)`.

gschaffner:
Closes #275.

Note that `inspect.signature` (which is used in the current `_invoke_with_optional_kwargs`!) is not available in Python < 3.3, which (I believe?) schema hasn't officially dropped support for yet. I know that I am just an outsider to this project, but I'd really suggest adopting #273.

Also, I think that my implementation below is not the most elegant way to deal with this issue. It might be better to instead have `Optional` accept a `pass_kwargs` parameter rather than choosing whether or not to pass `kwargs` based on a heuristic. `pass_args` could be `True` by default (to pass all kwargs), but could also be `False` or a set of strings (representing parameter names to pass). Let me know what you think!


skorokithakis:
Thanks for this! Yeah, 3.3 is ancient and wasn't very widely used, so I'd be in favor of dropping support for it. Would that make your PR easier/more elegant?

schema.py
```diff
@@ -274,10 +273,17 @@ def _priority(s):
...
-        return f()
-    return f(**kwargs)
+    try:
+        return f(**kwargs)
+    except TypeError as e:
+        if (
+            e.args[0].startswith("{}() got an unexpected keyword argument".format(f.__name__))  # Python 2/3.
+            or e.args[0].startswith("{}() takes no arguments".format(f.__name__))  # Python 2 only.
+        ):
+            return f()
```
BvB93, line 283: Is there a realistic scenario wherein someone should be passing `**kwargs` if the function does not support keyword-arguments in the first place? 
Why not just skip the whole `try/except` block and unconditionally return `f(**kwargs)`?

gschaffner: E.g. see `test_optional_callable_default_ignore_inherited_schema_validate_kwargs`: https://github.com/keleshev/schema/blob/09c00eda9599e53f7e6b84d7c91ecd3b42f71772/test_schema.py#L758-L770 (added in #268).

That approach would work if `lambda: 42` was instead `lambda **_: 42`. This would be a backwards-incompatible change (warranting a major version bump), though.

gschaffner:
Sorry for the delayed response!

Following @BvB93's comment above, I think the underlying problem here is how #268 passes around kwargs. Schema is trying to pass `validate`'s kwargs to the `default` argument of any `Optional` whose `default` accepts any keyword arguments. Ignoring the `inspect.signature(set/etc.)` issue for a moment, I don't entirely understand the rationale behind this choice. It means that if one wants to write a `default` argument that takes any kwargs, it also needs to accept (and ignore) extra kwargs intended for other `default` callables. This is a bit cumbersome when writing `default` inline, e.g. `Optional(default=lambda of_interest, **_: of_interest + 1)`.

Given this, though, `_invoke_with_optional_kwargs(f, **kwargs)` could be replaced unconditionally with `f(**kwargs)` if Schema adds the requirement that `default` arguments of `Optional`s must accept arbitrary kwargs. This would look like changing `lambda: 42` to `lambda **_: 42` in `test_optional_callable_default_ignore_inherited_schema_validate_kwargs`.

This approach is fundamentally flawed for other reasons, though. For example, suppose `Optional(default=dict)`. Now `validate(some_kwarg=...)` will cause the `dict` init to be `dict(some_kwarg=...)`! This is unexpected and very confusing behavior for users trying to use the `validate(**kwargs)` feature; they would instead need to use `Optional(default=lambda **_: dict())`. Also note that this `dict(some_kwarg=...)` issue is not a problem that can be fixed by simply detecting if `issubclass(f, dict)`; e.g. consider (a) non-stdlib `typing.Mapping` types that do not inherit from `dict`, or (b) `dict` subclasses with constructors that don't accept `**kwargs` but do accept a non-arbitrary list of kwargs that we'd like to pass via `validate()`, or (c) in general constructors that accept *and use* arbitrary kwargs.

Now, if #268 is to be kept and not rolled back, I think that the best way to deal with this is to (a) drop Python < 3.3 so that `inspect.signature` can be used (see #273) (or use an `inspect` backport for old Pythons) and (b) make `_invoke_with_optional_kwargs(f, **kwargs)` inspect the signature of `f` and pass only the subset of the kwargs that match `f`'s signature. If `f` is a builtin that `inspect.signature` fails on, `_invoke_with_optional_kwargs` should simply call it without arguments.

This would be a backwards-incompatible change that would necessitate a major version bump, but AFAIK it elegantly handles all of the relevant issues here: `Optional(default=set/etc.)`, `dict(some_kwargs=...)`, `lambda **kwargs: f"Hello, {kwargs['name']}"` rather than `lambda name: f"Hello, {name}"`.

I've implemented this below and adjusted the tests relevant tests that were added in #268 accordingly.

gschaffner:
On second thought: a major version bump is *technically* not necessary under SemVer even if this is merged, since Schema is still in major version 0. Schema still being v0.x.x means that if the major version is bumped, the reason for that bump should be to indicate public API stability, not to indicate backwards-incompatible changes.

I am under the impression that the public API has already been stable for some time, though, so perhaps the major version should still be bumped if this is merged.

PR #278, opened 2022-06-01
Type stub

eggplants:
To use this package with mypy in [jjjake/internetarchive](https://github.com/jjjake/internetarchive), I have to add `schema.pyi`.


skorokithakis:
Thanks for this! This is a pretty big PR, and it changes multiple parts, so it might take a while to review. For the future, it would be better if PRs were split up into self-contained bits that were easier to review.

eggplants:
splitted.

PR #279, opened 2022-06-01
Add some linting hooks in pre-commit

eggplants:
ex: #278


skorokithakis:
Hmm I ran this locally and pre-commit fails, which means that any commit will fail...

eggplants:
- setup-cfg-fmt
  - for: #280
- mypy
  - for: #278

skorokithakis:
Hmm, what is the oldest version that PyUpgrade made this compatible with?

PR #301, opened 2023-09-25
Improved Or class comment to document only_one purpose.

juliendelplanque:
Hello,

As part of a coding dojo we organized in our team, we took the opportunity to see what could be improved in this library (which is really nice by the way!). Here, we just slightly improved Or class comment.

Cheers,

Julien


skorokithakis (approved):
Thanks!

PR #304, opened 2023-11-07
Readme code samples readability

sezanzeb:
Some of the code examples were extremely hard to read.

For the most part I just used black to format them, except in the smaller examples, in which I put the curly braces and parenthesis on the same line to make it a bit more compact.

For the first example I omitted the `>>>` and `...` characters to make it easier to copy and paste it to try it locally.


skorokithakis (approved):
This looks good, thanks!

PR #305, opened 2023-11-08
Make mock test dependency optional for Python > 3.3

jelly:
Since Python 3.3 unittest.mock is available as a build in module.

Tox still mentions mock, but I have no idea how to make that optional.


skorokithakis (approved):
Thanks!

PR #319, opened 2024-06-09
fix: Include py.typed module when building package

### LINKED ISSUE #316: py.typed missing in package
I see that `schema.py` is typed and it would be great to be able to use that information when running mypy. In order for that to happen, there needs to be a `py.typed` file in the package.

I see that there was an attempt in including it in the package:
https://github.com/keleshev/schema/blob/df0d03f80554ea5de55628be4749252847c501eb/setup.py#L26
It seems to do nothing though when inspecting the `wheel` and `tar.gz` (version 0.7.7): only the `schema.py` file is stored.

I did not try but it might work if the file structure is changed to something like this:
```
schema
├── __init__.py <- current content of schema.py
└── py.typed
```

skorokithakis:
This is actually @maximeflya's #317, which fixes #316. I realized the formatting CI action was wrong (it should run pre-commit), and thus #317 was needlessly reformatted. I created this PR to hopefully get things right.

@maximeflya, does this look ok?


maximeflya:
You probably should clean your linters config files:
In `pyproject.toml`, you define the configuration for `black`
In `setup.cfg`, you define the configuration for `isort` and `flake8` 
And in the pre-commit hook, you run `ruff` which is not configured anywhere 

In my PR the formatting changed because black is configured to a line length of 120 (also flake8) where its default is 88. (Personally I prefer keeping the default)

In this case though, I don't think your CI is doing anything:
```
[INFO] Initializing environment for https://github.com/astral-sh/ruff-pre-commit.
[INFO] Installing environment for https://github.com/astral-sh/ruff-pre-commit.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
ruff.................................................(no files to check)Skipped
ruff-format..........................................(no files to check)Skipped
```
Maybe there is a parameter to run it on the whole code base in the CI.

Overall running the pre-commit hook like you did is very clean if it works :+1: I wasn't familiar with `pre-commit` but that's kind of nice 

(Also I'm not sure what the `semantic release` parameter is used for in the `pyproject.toml`, but it would be incorrect with the new changes.

skorokithakis:
Thank you! I cleaned stuff up in setup.cfg and pyproject.toml, as per your suggestion! I had assumed that the pre-commit action will be running on all files (as there's no point otherwise), but I've forced it to now.

maximeflya (approved):

Issue #318, opened 2024-06-04
Needless call to eval in setup.py


moltenform:
I prefer to get rid of calls to `eval` for security reasons,

Especially in a setup.py file where one wouldn't expect code to run just by importing the file.

If someone malicious were to make a file named `schema.py` they could get an arbitrary line of code to run.

Let's see if we can change the line 

```
 if line.startswith("__version__ ="):
            version = eval(line.split("=", 1)[1])
            break
```

to
 
```
 if line.startswith("__version__ ="):
            version = line.split("=", 1)
            version = version.replace('"', '').replace("'", '')
            version = version.strip()
            break
```

skorokithakis:
No reason why we can't, would you like to submit a PR?

moltenform:
Thanks!

skorokithakis:
Thank you!

PR #320, opened 2024-06-10
More straightforward way to retrieve the current version number

moltenform:
Related to
https://github.com/keleshev/schema/issues/318
, in my opinion this is safer than an eval call.


skorokithakis (approved):
Thanks!

skorokithakis (approved):

PR #330, opened 2025-02-20
JSON Schema: Add title for Literal, ECMA regexes, and null type in const

dblanchette:
- feat: [JSON Schema] Add title for Literal
- fix: [JSON Schema] attempt to make regexes ECMA-compliant
- fix: [JSON Schema] Use "type": "null" instead of "const": "null"


skorokithakis (approved):

skorokithakis:
Looks good to me, thanks!

PR #337, opened 2025-10-16
[Hotfix] Make Use class callable to resolve mypy type errors with And/Or 

### LINKED ISSUE #336: Type incompatibility introduced in version 0.7.8
After upgrading to Schema 0.7.8, type-checking tools (e.g., mypy) report new incompatibilities when using the And combinator with Use. The issue did not occur in previous versions (≤ 0.7.7).

```
error: Argument 2 to "And" has incompatible type "Use"; expected "Callable[..., Any]"  [arg-type]
```

steps to reproduce

```python
from schema import And, Use

# Minimal example
schema = And(int, Use(int))

```

Running mypy on this snippet with Schema 0.7.8 triggers the above error.

qequ:
When using `Use` with `And` or `Or` combinators, mypy reported a type error:

```python
from schema import And, Use

schema = And(int, Use(int))
```

**Error:**
```
Argument 2 to "And" has incompatible type "Use"; expected "Callable[..., Any]"
```

## Solution

Added `__call__` method to the `Use` class, making it properly callable:

```python
def __call__(self, data: Any) -> Any:
    return self._callable(data)
```

Closes #336

### DIFF
--- a/test_mypy_use_callable.py
+++ b/test_mypy_use_callable.py
@@ -3,24 +3,26 @@
 This test actually runs mypy as a subprocess and verifies that the specific
 error about Use being incompatible with And/Or is NOT present.
 """
+
+import shutil
 import subprocess
 import sys
 import tempfile
 from pathlib import Path
-import shutil
+
 import pytest
 
 
 # Check if mypy is available
 def is_mypy_available():
     """Check if mypy is available in the environment."""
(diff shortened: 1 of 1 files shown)


skorokithakis (approved):

skorokithakis:
Makes sense, thank you!

PR #343, opened 2026-03-04
Fix typing for `And` and `Or`

### LINKED ISSUE #342: Wrong type annotation for `And` and `Or`
The constructors of `And` and `Or` are typed as `__init__(self, *args: Union[TSchema, Callable[..., Any]], ...)`.

However, each `s in *args` is converted to a `Schema` automatically in `_build_schemas` which ultimately constructs a `Schema` object for each `s`. Therefore, `*args` should have the same annotation as `Schema.__init__(self, schema: Any, ...)`.

Moreover, [README](README.rst) says the following:

> #### Enumerated values
> 
> Use Or.
> 
> Example:
> 
>     Schema(Or(1, 2, 3))

But following this currently results in an `Argument of type "Literal[1]" cannot be assigned to parameter "args" of type "TSchema@Or | ((...) -> Any)"` error.

moi90:
Fixes #342.

### DIFF
--- a/schema/__init__.py
+++ b/schema/__init__.py
@@ -131,7 +131,7 @@ class And(Generic[TSchema]):
 
     def __init__(
         self,
-        *args: Union[TSchema, Callable[..., Any]],
+        *args: Any,
         error: Union[str, None] = None,
         ignore_extra_keys: bool = False,
         schema: Union[Type[TSchema], None] = None,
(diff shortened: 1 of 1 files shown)


skorokithakis (approved):

moi90:
Thanks for merging :partying_face:

skorokithakis:
Thanks for contributing!

